PR #108, opened 2019-09-13
WIP Remove runc binary building, add runc.io dependency

seemethere:
We can successfully build/release binaries now for `runc.io`! 

This is more of a POC to show that we can actually do this and not necessarily something we should merge in just yet.


seemethere:
This can probably be closed

PR #144, opened 2020-03-23
Force header check to run on amd64

thaJeztah:
Jenkins now has s390x machines enabled, which don't currently
have a CentOS 7 image on Docker Hub, causing the check to fail.


thaJeztah:
ping @silvin-lubecki @StefanScherer PTAL

PR #141, opened 2020-03-19
Add new v1 and v2 runc shim binaries.

cpuguy83:
I tested the rpm but am not sure how to build the deb, at least not using the main makefile.
`make all` just builds the rpm, `make deb` says no such target.

PR #124, opened 2020-01-09
Update packaging to support 1.3.2

ehazlett:
Tested so far on Ubuntu (I have to get some RPM distros up still).

```
$> apt install -y ./containerd.io_1.3.2-1_amd64.deb 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'containerd.io' instead of './containerd.io_1.3.2-1_amd64.deb'
The following NEW packages will be installed:
  containerd.io
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/25.0 MB of archives.
After this operation, 118 MB of additional disk space will be used.
Get:1 /root/build/ubuntu/eoan/amd64/containerd.io_1.3.2-1_amd64.deb containerd.io amd64 1.3.2-1 [25.0 MB]
Selecting previously unselected package containerd.io.
(Reading database ... 62528 files and directories currently installed.)
Preparing to unpack .../containerd.io_1.3.2-1_amd64.deb ...
Unpacking containerd.io (1.3.2-1) ...
Setting up containerd.io (1.3.2-1) ...
Created symlink /etc/systemd/system/multi-user.target.wants/containerd.service → /lib/systemd/system/containerd.service.
```

```
$> containerd --version
containerd containerd.io 1.3.2 ff48f57fc83a8c44cf4ad5d672424a98ba37ded6
```

```
Package: containerd.io
Status: install ok installed
Priority: optional
Section: devel
Installed-Size: 115562
Maintainer: Containerd team <help@containerd.io>
Architecture: amd64
Version: 1.3.2-1
Replaces: containerd, runc
Provides: containerd, runc
Depends: libc6 (>= 2.14), libseccomp2 (>= 2.4.0)
Conflicts: containerd, runc
Conffiles:
 /etc/containerd/config.toml 417033b466af4a9c58e3fc5842303008
Description: An open and reliable container runtime
Description-md5: b760064be8ea75804c76adda8f7a6c49
Homepage: https://containerd.io
```

```
$> ctr run -t --rm docker.io/library/busybox:latest shell echo ohai
ohai
```

PR #146, opened 2020-03-23
build all binaries instead of individual targets

thaJeztah:
This is an alternative approach for https://github.com/docker/containerd-packaging/pull/141 and the packaging changes in https://github.com/docker/containerd-packaging/pull/124

closes https://github.com/docker/containerd-packaging/pull/141

This allows upstream containerd to define all binaries that
should be built.

With this change, on the release/1.2 branch, the following binaries
are built:

    bin/ctr
    bin/containerd
    bin/containerd-stress
    bin/containerd-shim
    bin/containerd-shim-runc-v1

And on the release/1.3 branch (and on master), the following binaries
are built:

    bin/ctr
    bin/containerd
    bin/containerd-stress
    bin/containerd-shim
    bin/containerd-shim-runc-v1
    bin/containerd-shim-runc-v2

Effectively, compared to the current build-scripts, these additional
binaries are included:

    bin/containerd-stress
    bin/containerd-shim-runc-v1
    bin/containerd-shim-runc-v2 (for v1.3.0 and above)

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -15,34 +15,37 @@
 #   limitations under the License.
 
 INSTALL_DIR=debian/containerd.io
-CONTAINERD_BINARIES=bin/containerd bin/containerd-shim bin/ctr
 
 %:
 	dh $@ --with systemd
 
 # GO_SRC_PATH and PACKAGE are defined in the dockerfile
 # VERSION and REF are defined in scripts/build-deb
-bin/%: ## Create containerd binaries
-	@echo "+ make -C $(GO_SRC_PATH) --no-print-directory VERSION=$${VERSION} REVISION=$${REF} PACKAGE=$${PACKAGE} $@"
-	@make -C $(GO_SRC_PATH) --no-print-directory VERSION=$${VERSION} REVISION=$${REF} PACKAGE=$${PACKAGE} $@
-	mkdir -p $(@D)
-	mv -v $(GO_SRC_PATH)/$@ $@
+binaries: ## Create containerd binaries
+	@set -x; make -C $(GO_SRC_PATH) --no-print-directory \
+		DESTDIR="$$(pwd)" \
+		VERSION=$${VERSION} \
+		REVISION=$${REF} \
+		PACKAGE=$${PACKAGE} \
+		binaries install
+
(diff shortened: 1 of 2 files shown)


thaJeztah:
@crosbymichael @cpuguy83 PTAL; wondering if `containerd-stress` should be included or not

thaJeztah:
I'll change this to exclude the containerd-stress binary from the packages

thaJeztah:
@cpuguy83 @silvin-lubecki @ehazlett - updated this one to exclude `containerd-stress` from the packages

seemethere:
I originally didn't do this approach since it isn't really explicit in terms of what binaries end up in the package at then end.

thaJeztah:
Ah, yes, that makes sense as well. My train of thought with this change was to let upstream (where possible) determine what binaries containerd consists of (and this repo only wrapping them in a package), hopefully leading to less maintenance here, and allowing the same scripts to be used for any branch.

thaJeztah:
@seemethere now that you're here: do you recall if there's a specific reason we `.tar.gz` an archive to publish as Jenkins artefact, and not push the `.deb` / `.rpm` directly? If there's no specific reason, I'm thinking of removing that part; `.deb` and `.rpm`'s are already compressed, so it doesn't safe space, and it adds some complexity that possibly isn't needed.

debian/rules
```diff
@@ -14,35 +14,36 @@
...
+binaries: ## Create containerd binaries
+	@set -x; make -C $(GO_SRC_PATH) --no-print-directory \
+		DESTDIR="$$(pwd)" \
+		VERSION=$${VERSION} \
+		REVISION=$${REF} \
+		PACKAGE=$${PACKAGE} \
+		binaries install
+
+	# Remove containerd-stress, as we're not shipping it as part of the packages
+	rm -f bin/containerd-stress
```
silvin-lubecki, line 31: This one is weird, we should update `install` target upstream 🤔

thaJeztah: Yeah, was considering that, but likely wouldn't be back ported to existing release branches, so thought this was the "safer" option for now

PR #149, opened 2020-03-24
deb: use non-interactive, and no-install-recommends

thaJeztah:
The devscripts package on newer Ubuntu/Debian versions adds a HUGE
amount of new dependencies. Unfortunately we need this package as
it's the only package that provides `mk-build-deps`, but we don't
have to install all recommended packages:

On Ubuntu 20.04, without `--no-install-recommends`:

    4 upgraded, 334 newly installed, 0 to remove and 4 not upgraded.
    Need to get 100 MB of archives.
    After this operation, 445 MB of additional disk space will be used.

And with `--no-install-recommends`;

    0 upgraded, 104 newly installed, 0 to remove and 8 not upgraded.
    Need to get 25.0 MB of archives.
    After this operation, 119 MB of additional disk space will be used.

This patch also sets `DEBIAN_FRONTEND=noninteractive`, to prevent build
hanging if a package (such as `tzdata`) waits for user input otherwise.

The second commit updates handling of apt package cache; This patch removes the apt-cache after installing packages, and refreshes (`apt-get update`) before installing dependencies using `mk-build-deps` to prevent using a possibly stale index when dependencies in the control file change.

### DIFF
--- a/dockerfiles/deb.dockerfile
+++ b/dockerfiles/deb.dockerfile
@@ -42,7 +42,14 @@ RUN cat /etc/os-release
 ARG DEBIAN_FRONTEND=noninteractive
 
 # Install some pre-reqs
-RUN apt-get update && apt-get install -y --no-install-recommends curl devscripts equivs git lsb-release
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl \
+    devscripts \
+    equivs \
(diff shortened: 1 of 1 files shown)


thaJeztah:
bringing this one in 👍

PR #150, opened 2020-03-24
deb: add missing format version

thaJeztah:
dpkg-source: warning: no source format specified in debian/source/format, see dpkg-source(1)
    dpkg-source: warning: source directory 'containerd' is not <sourcepackage>-<upstreamversion> 'containerd.io-1.2.13'
    dpkg-source: info: using source format '1.0'


thaJeztah:
FWIW: I looked into using the recommended `3.0 (native)` or `3.0 (quilt)` formats (https://manpages.debian.org/testing/dpkg-dev/dpkg-source.1.en.html), but those require quite some changes to our packaging. We currently don't publish source packages, so for now probably not worth spending too much time on.

PR #153, opened 2020-03-25
Jenkinsfile: show information about CI environment

thaJeztah:
splitting this of from https://github.com/docker/containerd-packaging/pull/151 as it's useful to have information about the environment in which CI is running (for debugging purposes)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,8 +29,13 @@ def generatePackageStep(opts, arch) {
         node("linux&&${arch}") {
             stage("${opts.image}-${arch}") {
                 try {
-                    checkout scm
+                    sh 'docker version'
+                    sh 'docker info'
+                    sh '''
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -29,8 +29,13 @@ def generatePackageStep(opts, arch) {
         node("linux&&${arch}") {
             stage("${opts.image}-${arch}") {
                 try {
-                    checkout scm
+                    sh 'docker version'
+                    sh 'docker info'
+                    sh '''
```
silvin-lubecki: nit: this item is displayed as "Shell script" in the Jenkins logs, which is not very descriptive... Is there a way to add something like a title ?

thaJeztah: I think there's a third-party plugin to add labels, but don't think it's in Jenkins code (looked at that in moby/moby, but we decided not to use it and stick to core functionality where possible)

PR #154, opened 2020-03-26
Refactor to build in Dockerfile

thaJeztah:
Lots of changes, so best to look at each individual commit's description

PR #156, opened 2020-03-27
Fix some linting and formatting issues in scripts

thaJeztah:
extracting this from https://github.com/docker/containerd-packaging/pull/154

use https://github.com/docker/containerd-packaging/pull/156/files?w=1 to view the diff with whitespace-only changes hidden

### DIFF
--- a/scripts/.rpm-helpers
+++ b/scripts/.rpm-helpers
@@ -1,43 +1,43 @@
 #!/usr/bin/env bash
 
-
 gen-test-ver() {
-    local version=$1
-    local suffix=$2
-    local base_release_ver=$3
-    SPEC_FILE=${SPEC_FILE:-SPECS/containerd.spec}
-    test_num=${version#*-${suffix}.}
-    without_rpm_ver="${base_release_ver}.${test_num}.${suffix}.${test_num}"
-    # This only really matters for blessed builds (i.e. builds that are only rpm specific or have a version specific in the spec)
-    # should output the latest rpm build number so that it gets placed accordingly in the repo
-    if grep -q "${version%-${suffix}.*}-${without_rpm_ver}" "${SPEC_FILE}"; then
-        grep -o "${version%-${suffix}.*}-${without_rpm_ver}.[0-9]*" "${SPEC_FILE}" | head -n 1 | cut -d"-" -f 2
-    else
-        echo "${without_rpm_ver}.1"
-    fi
+	local version=$1
+	local suffix=$2
+	local base_release_ver=$3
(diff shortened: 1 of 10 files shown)


thaJeztah:
ping @tiborvass @ehazlett @silvin-lubecki  PTAL

scripts/.rpm-helpers
```diff
@@ -1,43 +1,43 @@
...
-    without_rpm_ver="${base_release_ver}.${test_num}.${suffix}.${test_num}"
-    # This only really matters for blessed builds (i.e. builds that are only rpm specific or have a version specific in the spec)
-    # should output the latest rpm build number so that it gets placed accordingly in the repo
-    if grep -q "${version%-${suffix}.*}-${without_rpm_ver}" "${SPEC_FILE}"; then
-        grep -o "${version%-${suffix}.*}-${without_rpm_ver}.[0-9]*" "${SPEC_FILE}" | head -n 1 | cut -d"-" -f 2
-    else
-        echo "${without_rpm_ver}.1"
-    fi
+gen_test_ver() {
+	version=$1
```
tiborvass: is there a linter that prefers global variables over local ?

thaJeztah: local is only supported on bash, so could fail when using in a Dockerfile

thaJeztah: I could use a sub-shell for this, but I don't think it was needed in this case

thaJeztah:
@tiborvass updated; ptal

ehazlett:
LGTM

scripts/.rpm-helpers
thaJeztah: ^ updated the functions to use sub-shells so that setting environment variables in a function does not change the current environment

scripts/new-rpm-release
```diff
@@ -18,20 +18,22 @@ set -e
 
 SPEC_FILE="rpm/containerd.spec"
 export SPEC_FILE
-source scripts/.rpm-helpers
+
+# shellcheck disable=SC1090
+source "$(dirname "$0")/.rpm-helpers"
```
silvin-lubecki: I guess it's `. "$(dirname "$0")/.rpm-helpers"` instead?

thaJeztah: Ah, yes, could update it; this one still required Bash though, so there's still other things that are non-POSIX

thaJeztah: Oh, actually; looks like those were no longer needed; let me update 👍

thaJeztah: Updated 👍

PR #159, opened 2020-03-27
rpm: fix Oracle Linux 8 build

thaJeztah:
extracting this from https://github.com/docker/containerd-packaging/pull/154

the extra repo was set unconditionally, which caused
Oracle Linux 8 builds to fail, because it doesn't have
that repo.


thaJeztah:
ping @tiborvass @ehazlett @silvin-lubecki PTAL

Issue #69, opened 2018-12-13
man pages for deb packages


andrewhsu:
The man pages are available for `rpm` packages:
```
$ rpm -ql containerd.io
/etc/containerd
/etc/containerd/config.toml
/usr/bin/containerd
/usr/bin/containerd-shim
/usr/bin/ctr
/usr/lib/systemd/system/containerd.service
/usr/sbin/runc
/usr/share/doc/containerd.io-1.2.1
/usr/share/doc/containerd.io-1.2.1/README.md
/usr/share/licenses/containerd.io-1.2.1
/usr/share/licenses/containerd.io-1.2.1/LICENSE
/usr/share/man/man1/containerd-config.1
/usr/share/man/man1/containerd.1
/usr/share/man/man1/ctr.1
/usr/share/man/man5/containerd-config.toml.5
```

But on part of the `deb` packages:
```
$ dpkg-query -L containerd.io
/.
/etc
/etc/containerd
/etc/containerd/config.toml
/lib
/lib/systemd
/lib/systemd/system
/lib/systemd/system/containerd.service
/usr
/usr/bin
/usr/bin/containerd
/usr/bin/containerd-shim
/usr/bin/ctr
/usr/sbin
/usr/sbin/runc
/usr/share
/usr/share/doc
/usr/share/doc/containerd.io
/usr/share/doc/containerd.io/changelog.gz
/usr/share/doc/containerd.io/copyright
```

Issue #4, opened 2018-05-01
package man pages


andrewhsu:
Need to be able to run `man ctr`. There are now targets to build man pages: https://github.com/containerd/containerd/pull/2079

PR #165, opened 2020-04-02
deb: include man pages

thaJeztah:
closes https://github.com/docker/containerd-packaging/pull/8
fixes https://github.com/docker/containerd-packaging/issues/4
fixes https://github.com/docker/containerd-packaging/issues/69


rpm/containerd.spec
```diff
@@ -121,11 +121,13 @@ install -D -m 0755 bin/* %{buildroot}%{_bindir}
 install -D -m 0644 %{S:1} %{buildroot}%{_unitdir}/containerd.service
 install -D -m 0644 %{S:2} %{buildroot}%{_sysconfdir}/containerd/config.toml
 
-# install manpages
-install -d %{buildroot}%{_mandir}/man1
-install -p -m 644 man/*.1 $RPM_BUILD_ROOT/%{_mandir}/man1
-install -d %{buildroot}%{_mandir}/man5
-install -p -m 644 man/*.5 $RPM_BUILD_ROOT/%{_mandir}/man5
+# install manpages, taking into account that not all sections may be present
+for i in $(seq 1 8); do
```
chris-crone, line 125: I see that we used to fill man 1 and 5 and that we're now doing 1 and 8, was this change intended?

thaJeztah: Yes; the man-pages were in the wrong section (1 instead of 8), this was fixed upstream, but now the packaging script here didn't take section 8 into account.

I changed it to take all sections (1..8) into account, so that we won't have to update this again if new man pages are added in future that are in different sections

PR #166, opened 2020-04-03
Makefile: follow upstream Go version by default

thaJeztah:
The master branch requires Go 1.13 or up, but other branches may still expect an older version of Go.

This patch makes the Go version used default to the version that's used in upstream, but allows overriding the version by manually setting the `GOVERSION` make variable.

Also setting a default value for the golang image in the Dockerfiles (defaulting to `golang:latest`)


thaJeztah:
ping @glours @silvin-lubecki PTAL

PR #169, opened 2020-04-20
Makefile: fix error if source directory does not yet exist

thaJeztah:


thaJeztah:
@silvin-lubecki @StefanScherer PTAL

StefanScherer:
Oh the failing centos:7 on armhf is caused by the EC2 agents. I'll make the Scaleway machines online again.
I have to look into that how to fix the build on an armv8l machine.

```
Cannot find a valid baseurl for repo: base/7/armv8l
```

thaJeztah:
ok, we should be able to ignore those then

PR #171, opened 2020-04-21
Remove CentOS armhf from matrix

thaJeztah:
The matrix of distro/architectures to build should match the list of docker packages. CentOS on armhf (arm32) was never supported, so let's remove it from the list.

Users are still able to build these packages if needed, but we don't have to run CI for it.


thaJeztah:
@silvin-lubecki @StefanScherer PTAL

thaJeztah:
Ok, this is green, except for CentOS on arm64, which is unrelated

StefanScherer:
Good idea. Belated LGTM 😅

PR #172, opened 2020-04-21
Makefile: allow overriding repository to build from

thaJeztah:
Our internal pipeline has a configuration to override the repository to build from. While it's not currently used, there could be situations where it's usefull (e.g. when building an embargoed security release).

This patch adds a CONTAINERD_REMOTE makefile variable that allows overriding the repository that we fetch the source from.

Also cleaning up the `CONTAINERD_MOUNT` variable, as it didn't serve any use


thaJeztah:
@silvin-lubecki ptal

PR #175, opened 2020-04-23
Remove "CREATE_ARCHIVE"; archive .deb/.rpm's artifacts directly

thaJeztah:
deb and rpm packages are already compressed, so compressing them again does not save any space.

This patch removes the "CREATE_ARCHIVE" option, and instead just the deb/rpm's itself as artifacts

### Before this

Tars are uploaded as artefact in Jenkins:

Screenshot 2021-11-22 at 11 48 17

### After this

The individual `.deb` and `.rpm` packages are uploaded as artefact:

Screenshot 2021-11-23 at 13 05 24

### DIFF
--- a/scripts/build-deb
+++ b/scripts/build-deb
@@ -62,8 +62,3 @@ if [ "${DIST_VERSION}" = 'n/a' ]; then
 	echo "Failed to get release codename"
 	exit 1
 fi
-
-# Only create an archive if env variable is specified
-if [ -n "${CREATE_ARCHIVE}" ]; then
-	tar -C /build -cvzf "/archive/${DIST_ID}-${DIST_VERSION}-${ARCH}.tar.gz" "${DIST_ID}/${DIST_VERSION}/${ARCH}"
-fi
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-archive/
 artifacts/
 build/
 common/containerd.service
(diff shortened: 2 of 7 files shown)


thaJeztah:
Need to double check if these archives are used elsewhere

@simonferquel @djs55 do you know if these are used by the Docker Desktop team?

djs55:
Looking at the build scripts I don't think Docker Desktop downloads `containerd` artefacts directly. We use the `linuxkit` `containerd` image and we download `dockerd` binary as a `.tar.gz` (which I think includes a `containerd` version?)

thaJeztah:
Looks like the release pipeline currently uses these tars to stash them in an s3 bucket (and later restores them). Need to figure out if there's a way to do this without having to use the intermediate `.tar.gz` files (challenge is likely that the tars are used to preserve the path / directory structure).

scripts/build-deb
```diff
@@ -62,8 +62,3 @@ if [ "${DIST_VERSION}" = 'n/a' ]; then
 	echo "Failed to get release codename"
 	exit 1
 fi
-
-# Only create an archive if env variable is specified
-if [ -n "${CREATE_ARCHIVE}" ]; then
-	tar -C /build -cvzf "/archive/${DIST_ID}-${DIST_VERSION}-${ARCH}.tar.gz" "${DIST_ID}/${DIST_VERSION}/${ARCH}"
-fi
```
thaJeztah, line 69: Note to self; instead of tarring here, perhaps we can use `docker build --output type=tar...` (although that doesn't support compressing, but I guess all artefacts are already compressed, so from that perspective, it should be fine)

thaJeztah:

PR #181, opened 2020-05-12
Add checkout.sh to handle tags with shallow clones

tiborvass:
Commit 9706347c34524feb83d75a7119ea88311b41e668 changed to shallow clones
but fetching a tag ref does not download the tag itself.
This patch makes sure that the tag (or branch) is downloaded otherwise
fallsback to fetching the commit hash as it did before.

Ref #173

### DIFF
--- a/scripts/checkout.sh
+++ b/scripts/checkout.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env sh
+
+#   Copyright 2018-2020 Docker Inc.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+
+#       http://www.apache.org/licenses/LICENSE-2.0
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


scripts/checkout.sh
```diff
@@ -0,0 +1,27 @@
...
+		FETCH_REF="$REF"
+	fi
+	git -C "$SRC" fetch --depth 1 origin "$FETCH_REF"
+	git -C "$SRC" checkout -q "$REF_CANON"
+}
+
+
+# Only execute checkout function above if this file is executed, not sourced from another script
+prog=checkout.sh # needs to be in sync with this file's name
+if [ "$(basename -- $0)" == "$prog" ]; then
```
tiborvass, line 25: This check is not needed here, but i thought this file could be reused (in other scripts as well) as it's basically circumventing a git limitation. I wish this beahvior was just a flag on git fetch or that git clone accepted commit hashes.

scripts/checkout.sh
```diff
@@ -0,0 +1,40 @@
...
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+checkout() {
+	set -ex
+	local SRC="$1"
```
thaJeztah, line 19: I think `local` is non-POSIX, so would require bash

could use a sub shell instead (change `{}` to `()`)

PR #184, opened 2020-07-08
update runc to v1.0.0-rc91

AkihiroSuda:
Release note: https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc91

PR #191, opened 2020-07-21
Makefile: automatically select default runc version based on vendor.conf

thaJeztah:
alternative to / supersedes https://github.com/docker/containerd-packaging/pull/184

closes https://github.com/docker/containerd-packaging/pull/184

### DIFF
--- a/common/common.mk
+++ b/common/common.mk
@@ -17,13 +17,18 @@
 CONTAINERD_REMOTE ?=https://github.com/containerd/containerd.git
 RUNC_REMOTE       ?=https://github.com/opencontainers/runc.git
 REF?=HEAD
-RUNC_REF?=dc9208a3303feef5b3839f4323d9beb36df0a9dd
 
+# Select the default version of Golang and runc based on the containerd source.
+# The runc version commit/tag is defined in vendor.conf. Code below is based on
+# the code used in the containerd repository:
+# https://github.com/containerd/containerd/blob/499fbb0337c9138b5360117e0b25a7a1428f9667/script/setup/install-runc#L24
 ifdef CONTAINERD_DIR
 GOVERSION?=$(shell grep "ARG GOLANG_VERSION" $(CONTAINERD_DIR)/contrib/Dockerfile.test | awk -F'=' '{print $$2}')
+RUNC_REF?=$(shell grep "opencontainers/runc" "$(CONTAINERD_DIR)/vendor.conf" | awk '{print $$2}')
 else
-# TODO adjust GOVERSION macro to take CONTAINERD_REMOTE into account
+# TODO adjust GOVERSION and RUNC_REF macro to take CONTAINERD_REMOTE into account
 GOVERSION?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/contrib/Dockerfile.test" | grep "ARG GOLANG_VERSION" | awk -F'=' '{print $$2}')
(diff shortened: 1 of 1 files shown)


thaJeztah:
@akihirosuda @tiborvass @silvin-lubecki PTAL

AkihiroSuda:
containerd may vendor untagged version of libcontainer.

Also, we are likely to migrate to go.mod

thaJeztah:
> containerd may vendor untagged version of libcontainer.

I think for released versions, they'd be vendoring tagged versions, or do you expect that not to be the case?

> Also, we are likely to migrate to go.mod

I guess it'd be possible to support both. Hopefully runc reaches `v1.x` soon, so that we can get rid of this logic altogether.

Note that this is the _default_ value, but it can still be overridden when running `make`;

```bash
make RUNC_REF=v1.0.0-rc8 docker.io/library/ubuntu:focal

# Building packages on docker.io/library/ubuntu:focal
# 
# containerd   : HEAD (commit: ec05460)
# runc         : v1.0.0-rc8 (commit: 425e105)
# architecture : x86_64
# build image  : docker.io/library/ubuntu:focal
# golang image : docker.io/library/golang:1.13.14-buster
# --------------------------------------------------------------------
#
# ...
```

AkihiroSuda:
> I think for released versions, they'd be vendoring tagged versions, or do you expect that not to be the case?

We try to vendor tagged versions, but we don’t guarantee.

thaJeztah:
> We try to vendor tagged versions, but we don’t guarantee.

Right, but the version in `vendor.conf` is (at least currently, and at least until runc reaches "stable") still leading, correct? Basically, my train of thought with this change is that;

- getting the version from `vendor.conf` matches what upstream containerd/containerd is using (and testing against)
- it allows the scripts in this repository to be used against all of the upstream branches (and pick the version of runc matching that branch)
- effectively; making sure that the default "just works®"
- but allow that default to be overridden when building packages (so if we want to pin to a released version, we can set the `RUNC_REF` parameter)

common/common.mk
```diff
@@ -17,13 +17,18 @@
...
+# the code used in the containerd repository:
+# https://github.com/containerd/containerd/blob/499fbb0337c9138b5360117e0b25a7a1428f9667/script/setup/install-runc#L24
 ifdef CONTAINERD_DIR
 GOVERSION?=$(shell grep "ARG GOLANG_VERSION" $(CONTAINERD_DIR)/contrib/Dockerfile.test | awk -F'=' '{print $$2}')
+RUNC_REF?=$(shell grep "opencontainers/runc" "$(CONTAINERD_DIR)/vendor.conf" | awk '{print $$2}')
 else
-# TODO adjust GOVERSION macro to take CONTAINERD_REMOTE into account
+# TODO adjust GOVERSION and RUNC_REF macro to take CONTAINERD_REMOTE into account
 GOVERSION?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/contrib/Dockerfile.test" | grep "ARG GOLANG_VERSION" | awk -F'=' '{print $$2}')
+RUNC_REF?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/vendor.conf" | grep "opencontainers/runc" | awk '{print $$2}')
```
tiborvass, line 31: I think this is fine, but hopefully we won't need to support forks :)

tiborvass:
I guess we'll need to update this if containerd moves to go.mod

common/common.mk
thaJeztah: In that case we can manually set it (see above)

PR #192, opened 2020-08-13
Update systemd unit-file with changes from upstream

thaJeztah:
Includes upstream changes from;

- https://github.com/containerd/containerd/pull/3741 Add local-fs.target to service file
    - fixes https://github.com/containerd/containerd/issues/3671 Unable to recover corrupt image after unexpected host reboot
- https://github.com/containerd/containerd/pull/3297 add Restart=always to unit file
    - also added `RestartSec`, which was added in https://github.com/containerd/containerd/commit/1c7312e5da1ae8ca3cf238397939ed1d24f60758

We should change the packaging to use the upstream systemd unit, but looks like the path used in upstream is `/usr/local/bin/containerd` whereas packaging here uses `/usr/bin/containerd`, so I'll look at that separately.

### DIFF
--- a/common/containerd.service
+++ b/common/containerd.service
@@ -15,7 +15,7 @@
 [Unit]
 Description=containerd container runtime
 Documentation=https://containerd.io
-After=network.target
+After=network.target local-fs.target
 
 [Service]
 ExecStartPre=-/sbin/modprobe overlay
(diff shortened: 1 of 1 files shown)


thaJeztah:
@tianon @akihirosuda @silvin-lubecki PTAL

PR #194, opened 2020-08-13
Change install location from /usr/bin to /usr/local/bin

thaJeztah:
This changes the install location of packages from `/usr/bin` to `/usr/local/bin`
to match upstream.

Given that we're building a non-distro package, `/usr/local/bin` looks to be the
correct location; https://unix.stackexchange.com/a/8658

> /usr/bin is for distribution-managed normal user programs.
> /usr/local/bin is for normal user programs not managed by the distribution
> package manager, e.g. locally compiled packages. You should not install them
> into /usr/bin because future distribution upgrades may modify or delete them
> without warning.

The specification does not call out "distro" and "non-distro" packages, but does
explain that `/usr/bin` can be overwritten by system upgrades, and `/usr/local/bin`
should therefore be preferred;

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html#purpose24

> # 4.9. /usr/local : Local hierarchy
> ## 4.9.1. Purpose
>
> The /usr/local hierarchy is for use by the system administrator when installing
> software locally. It needs to be safe from being overwritten when the system
> software is updated. It may be used for programs and data that are shareable
> amongst a group of hosts, but not found in /usr.
>
> Locally installed software must be placed within /usr/local rather than /usr
> unless it is being installed to replace or upgrade software in /usr.
>
> Software placed in / or /usr may be overwritten by system upgrades (though we
> recommend that distributions do not overwrite data in /etc under these
> circumstances). For this reason, local software must not be placed outside of
> /usr/local without good reason.

After this change, we should also be able to switch to using the upstream systemd unit file

### DIFF
--- a/rpm/containerd.spec
+++ b/rpm/containerd.spec
@@ -121,8 +121,8 @@ make -C /go/src/github.com/opencontainers/runc BINDIR=%{_topdir}/BUILD/bin BUILD
 
 %install
 cd %{_topdir}/BUILD
-mkdir -p %{buildroot}%{_bindir}
-install -D -m 0755 bin/* %{buildroot}%{_bindir}
+mkdir -p %{buildroot}%{_exec_prefix}/local/bin
+install -D -m 0755 bin/* %{buildroot}%{_exec_prefix}/local/bin
 install -D -m 0644 %{S:1} %{buildroot}%{_unitdir}/containerd.service
 install -D -m 0644 %{S:2} %{buildroot}%{_sysconfdir}/containerd/config.toml
 
@@ -149,7 +149,7 @@ done
 %files
 %license LICENSE
 %doc README.md
-%{_bindir}/*
+%{_exec_prefix}/local/bin/*
 %{_unitdir}/containerd.service
 %{_sysconfdir}/containerd
 %{_mandir}/man*/*
--- a/common/containerd.service
+++ b/common/containerd.service
@@ -19,7 +19,7 @@ After=network.target
 
 [Service]
 ExecStartPre=-/sbin/modprobe overlay
-ExecStart=/usr/bin/containerd
+ExecStart=/usr/local/bin/containerd
 KillMode=process
 Delegate=yes
 LimitNOFILE=1048576
--- a/debian/rules
+++ b/debian/rules
@@ -51,8 +51,10 @@ override_dh_systemd_start:
 	dh_systemd_start --restart-after-upgrade
 	sed -i 's/_dh_action=try-restart/_dh_action=restart/g' ./debian/containerd.io.postinst.debhelper
 
+override_dh_usrlocal: # Skip processing dh_usrlocal
+
 override_dh_auto_install: binaries bin/runc man
-	mkdir -p debian/containerd.io/usr/bin
-	install -D -m 0755 bin/* debian/containerd.io/usr/bin
+	mkdir -p debian/containerd.io/usr/local/bin
+	install -D -m 0755 bin/* debian/containerd.io/usr/local/bin
 	install -D -m 0644 /root/common/containerd.service debian/containerd.io/lib/systemd/system/containerd.service
 	install -D -m 0644 /root/common/containerd.toml    debian/containerd.io/etc/containerd/config.toml


rpm/containerd.spec
```diff
@@ -121,8 +121,8 @@ make -C /go/src/github.com/opencontainers/runc BINDIR=%{_topdir}/BUILD/bin BUILD
 
 %install
 cd %{_topdir}/BUILD
-mkdir -p %{buildroot}%{_bindir}
-install -D -m 0755 bin/* %{buildroot}%{_bindir}
+mkdir -p %{buildroot}%{_exec_prefix}/local/bin
+install -D -m 0755 bin/* %{buildroot}%{_exec_prefix}/local/bin
```
thaJeztah, line 125: Unfortunately looks like there's no RPM macro for `/usr/local/bin`  `%{_exec_prefix}` (or `%{_prefix}`  not sure which one was preferred ?) came closest https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/#_macros_for_paths_set_and_used_by_build_systems

thaJeztah:
@seemethere @silvin-lubecki @tianon @akihirosuda PTAL

AkihiroSuda:
I don't think this is correct, `/usr` is the right place for rpm/deb packages.

Also this may break SELinux distros running with an older version of containers-selinux policy

thaJeztah:
Note that the official Docker images already use this location;

```bash
docker run --rm docker:19.03.12 ls -la /usr/local/bin/
total 195528
drwxr-xr-x    1 root     root          4096 Jun 25 19:20 .
drwxr-xr-x    1 root     root          4096 May 29 14:20 ..
-rwxr-xr-x    1 1000     1000      32751272 Jun 22 15:50 containerd
-rwxr-xr-x    1 1000     1000       6012928 Jun 22 15:50 containerd-shim
-rwxr-xr-x    1 1000     1000      18194536 Jun 22 15:50 ctr
-rwxr-xr-x    1 1000     1000      61108403 Jun 22 15:50 docker
-rwxrwxr-x    1 root     root          1850 Jun 25 19:20 docker-entrypoint.sh
-rwxr-xr-x    1 1000     1000        708616 Jun 22 15:50 docker-init
-rwxr-xr-x    1 1000     1000       2928514 Jun 22 15:50 docker-proxy
-rwxr-xr-x    1 1000     1000      68878880 Jun 22 15:50 dockerd
-rwxrwxr-x    1 root     root           587 Jun 25 19:20 modprobe
-rwxr-xr-x    1 1000     1000       9600712 Jun 22 15:50 runc
```

thaJeztah:
> I don't think this is correct, /usr is the right place for rpm/deb packages.

Is it the correct path for both distro- and non-distro packages? (wondering)

If it's not correct, should we update containerd upstream systemd unit file (and other places where it's used)?

thaJeztah:
The SELinux profile is a good point (I looked in the containerd repository itself for the AppArmor profile). I see for containerd; https://github.com/containers/container-selinux/blob/746ea7a2072a281106c5e0e2f9148ff0442e4d6e/container.fc#L6-L7

```
/usr/bin/containerd.*		--	gen_context(system_u:object_r:container_runtime_exec_t,s0)
/usr/local/bin/containerd.*		--	gen_context(system_u:object_r:container_runtime_exec_t,s0)
```

Looks like the man-page needs an update to include that as well; https://github.com/containers/container-selinux/blob/ae0720d6ac32866b023babeff75e766870951f8f/container_selinux.8#L312

```
/usr/bin/crio.*, /usr/bin/lxc-.*, /usr/bin/lxd-.*, /usr/bin/ocid.*, /usr/sbin/crio.*, /usr/sbin/ocid.*, /usr/bin/docker.*, /usr/lib/docker/[^/]*plugin, /usr/libexec/lxc/.*, /usr/libexec/lxd/.*, /usr/bin/container[^/]*plugin, /usr/bin/containerd.*, /usr/local/bin/crio.*, /usr/libexec/docker/.*, /usr/local/sbin/crio.*, /usr/libexec/docker/docker.*, /usr/bin/lxc, /usr/bin/lxd, /usr/bin/crun, /usr/bin/runc, /usr/sbin/runc, /usr/bin/podman, /usr/bin/fuidshift, /usr/local/bin/crun, /usr/local/bin/runc, /usr/local/bin/podman, /usr/bin/docker-latest, /usr/bin/docker-current, /usr/bin/rhel-push-plugin, /usr/sbin/rhel-push-plugin
```


`docker` itself looks to be only in `/usr/bin` currently; perhaps we should add `/usr/local/bin` to that as well https://github.com/containers/container-selinux/blob/746ea7a2072a281106c5e0e2f9148ff0442e4d6e/container.fc#L5

AkihiroSuda:
> Is it the correct path for both distro- and non-distro packages? (wondering)

I believe correct. I've (almost) never seen a rpm/deb that uses `/usr/local/bin`.

thaJeztah:
@silvin-lubecki let's wait for the discussion above. As alternative, we could use the upstream systemd unit file, and do a `s#/usr/local/bin#/usr/bin#`

thaJeztah:
Opened https://github.com/docker/containerd-packaging/pull/195 as alternative; I think that one will cause less disruption (and reduces maintenance, so looks like the better option)

tianon:
Yeah, this one is definitely not right -- packages installing to `/usr/local` is frowned upon by the FHS (and by extension, Debian Policy) -- in Debian specifically, this would be considered a "policy violation" and a release-blocking bug (https://www.debian.org/doc/debian-policy/ch-opersys#site-specific-programs).

The `docker` image installs to `/usr/local` specifically because it's installing from a tarball (so it's a "site specific" install, not from a package).

rpm/containerd.spec
tianon: That's probably because RPMs aren't supposed to touch it per the FHS :sweat_smile:

thaJeztah:
I guess I got misled by the StackOverflow answer (which to some extend made sense). Anyway; the other one got merged, and is a simpler as well, so life is good 👍

PR #208, opened 2020-12-02
Makefile: switch to POSIX mode to handle errors

thaJeztah:
similar to https://github.com/docker/docker-ce-packaging/pull/510 and https://github.com/moby/sys/pull/19#discussion_r445698069

relates to https://github.com/docker/docker-ce-packaging/pull/504#issuecomment-737078670

> .POSIX
>
> The application shall ensure that this special target is specified without prerequisites
> or commands. If it appears as the first non-comment line in the makefile, make shall
> process the makefile as specified by this section; otherwise, the behavior of make is
> unspecified.

Running the makefile with `.POSIX` runs shells with the `-e` options, which helps with
handling errors; without this, make can complet "succesfully", even if shell commands
in a target fail.

Note that this patch does not change behavior on macOS, which runs an older version of
GNU make that does not support these options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+.POSIX:
+
 include common/common.mk
 
 BUILD_IMAGE=centos:7


thaJeztah:
@silvin-lubecki @tianon PTAL

silvin-lubecki (dismissed):
LGTM

thaJeztah:
Rebased to check if this catches the failure we currently have in automatic runc-version detection because containerd "master" switched to `go.mod`, therefore these lines now fail; https://github.com/docker/containerd-packaging/blob/484fe7cdeaeb2711cb0a823bff4d5ee51009f0ff/common/common.mk#L27-L31

thaJeztah:
ok, that still didn't catch that problem; we can probably have a look at another approach for that

thaJeztah:

PR #209, opened 2020-12-02
Force Go to 1.13

tiborvass:
Since Go 1.14, runtime changes make it possible for syscall functions
to return EINTR errors. We are not aware of an analysis of such callers.

As this could impact users badly and in obscure ways this patch ensures
containerd and runc binaries are built with Go 1.13 for now.


tiborvass:
Wrong repo

PR #215, opened 2021-01-21
remove default "containerd.toml" config file

thaJeztah:
This configuration file was used to disable the CRI plugin to work around an
issue where CRI was automatically listening for remote connections.

That problem was resolved, so we no longer need a customized configuration.

Addresses https://github.com/containerd/containerd/issues/4956


thaJeztah:
@cpuguy83  @oldthreefeng @tiborvass PTAL

cpuguy83 (dismissed):
LGTM

oldthreefeng:
LGTM

thaJeztah:
whoops; missed a reference to that file in the rpm spec; pushing a fix now

rpm/containerd.spec
```diff
@@ -151,9 +149,7 @@ done
 %doc README.md
 %{_bindir}/*
 %{_unitdir}/containerd.service
-%{_sysconfdir}/containerd
 %{_mandir}/man*/*
-%config(noreplace) %{_sysconfdir}/containerd/config.toml
```
thaJeztah, line 156: This is the only part I was looking at if this could have side-effects (file and directory no longer marked as "config"). I _think_ it should have no consequences (package no longer knows of this path, so won't interfere with it), but if someone knows if this is a concern, let me know.

thaJeztah:
Updated; and all green now; @cpuguy83 @tianon PTAL 🤗

tianon:
Nice! :+1:

Debian has a `conffiles` thing that might interfere here (https://www.debian.org/doc/manuals/maint-guide/dother.en.html#conffiles; by default it marks anything installed to `/etc` as a "conffile" which then needs extras to properly remove/manage).

We probably need a `<package>.maintscript` file that includes `rm_conffile /etc/containerd/config.toml` (https://manpages.debian.org/buster/debhelper/dh_installdeb.1.en.html, https://manpages.debian.org/buster/dpkg/dpkg-maintscript-helper.1.en.html).

thaJeztah:
> We probably need a <package>.maintscript file that includes rm_conffile /etc/containerd/config.toml (https://manpages.debian.org/buster/debhelper/dh_installdeb.1.en.html, https://manpages.debian.org/buster/dpkg/dpkg-maintscript-helper.1.en.html).

@tianon IIUC, you mean "add a script that removes the config file when installing the new package"? I wonder if we should do that; even though we provided a "default", we also marked the file as `noreplace`, which (IIUC) means that a user could have added their own custom configuration file there (in which case we wouldn't overwrite it). If we would do a "remove" of that file when installing the package, we would remove their custom config, right?

Or do you mean the "reverse"; prevent the default behavior of `.deb` from removing the file, now that we no longer list it (and mark it with `noreplace` ? (that was my initial concern with removing that line)

tianon:
Discussed in maintainers meeting with @thaJeztah @cpuguy83; the config file we were installing is very benign and shouldn't cause problems for any existing users if it doesn't get cleaned up (and the chance of breaking changes to that config file syntax are low), so this is purely a perfectionism / cleanliness concern and should be relatively harmless to ignore.

In other words, we do not _need_ to include `rm_conffile` and it should be fine to just leave alone and stop adding for new installs.

zhsj:
fwiw, rm_conffile in dpkg-maintscript-helper will not remove conffile if it has been modified by users. dpkg has md5sum of the conffile which is used when removing.

zhsj:
> fwiw, rm_conffile in dpkg-maintscript-helper will not remove conffile if it has been modified by users. dpkg has md5sum of the conffile which is used when removing.

ha, seems i misread the above discussion... (just to ignore that)

thaJeztah:
> ha, seems i misread the above discussion... (just to ignore that)

No worries; thanks for looking! Yes, we discussed for a bit, and thought it was not worth the extra complexity to clean up the file (which is mostly harmless)

thaJeztah:
@silvin-lubecki @tiborvass ptal; this one good to go?

silvin-lubecki (approved):
LGTM

thaJeztah:
@tiborvass ptal

fuweid (approved):
LGTM(nb)

thaJeztah:
Given that this fix is not critical, and it's been pending for a while, I think it should be ok to merge this after https://github.com/docker/containerd-packaging/pull/223 has been merged and released (keeping the diff for that release small, as it's a security release, so better to not include too many other changes)

thaJeztah:
Ah, yes, let's get this one in 👍

rpm/containerd.spec
ravilr: @thaJeztah removing this does have some impact on rpm users who are currently overwriting the rpm supplied config.toml with their own.  This being removed above, causes the user supplied override config.toml file to be deleted, when they upgrade to a rpm version with this change.  rpm moves the file `.rpmsave` and `.rpmnew` after installing the new version with above change.

Can we just continue shipping an empty config.toml with just `version=2` without the disabled cri plugin config ?

thaJeztah: @ravilr hmmm... yeah, we could partially revert that change if there's no other way; do you know if rpms allow for the change to be "ignored" (as in: don't touch the existing file?)

thaJeztah:
Opened https://github.com/docker/containerd-packaging/pull/233 to (temporarily) revert while we figure out an approach for https://github.com/docker/containerd-packaging/pull/215#discussion_r632804641

rpm/containerd.spec
ravilr: according to https://rpm-packaging-guide.github.io/#files,  %config(noreplace) exists to protect local modifications to a config file, so that local edits to config files won't get lost during a subsequent upgrade.   
But, since the file is being removed from rpm, that doesn't apply here. 

IMO,  if the intention is to get rid of disabling of cri plugin, then rpm packaging here should continue shipping an empty config.toml with %config(noreplace).

PR #218, opened 2021-02-19
deb: use "containerd.io" as workdir to suppress some warnings

thaJeztah:
dpkg-source expects the directory to match the package name (containerd.io)

    dpkg-source: warning: source directory 'containerd' is not <sourcepackage>-<upstreamversion> 'containerd.io-0.20210219.014044~e58be59'


thaJeztah:
oh, probably still doesn't work, because the directory doesn't have `-<upstreamversion>` 😅

PR #213, opened 2021-01-12
Jenkinsfile: test building Debian 11 "bullseye"

thaJeztah:
Depends on:

- [x] https://github.com/docker/containerd-packaging/pull/212 deb: make dh-systemd dependency optional as it's deprecated
- [x] https://github.com/docker/containerd-packaging/pull/219 deb: keep '/var/lib/apt/lists/' to allow building for Debian unstable

PR #219, opened 2021-02-19
deb: keep '/var/lib/apt/lists/' to allow building for Debian unstable

thaJeztah:
relates to https://github.com/docker/containerd-packaging/pull/213#issuecomment-782172567

Debian "unstable" releases use apt caching information to get the codename
see discussion on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651:

> That's all to say that this bug is (to my belief) actually expected behaviour;
> and fixing it through forcing the codename to be interpreted as "stretch" when
> apt-cache information is unavailable would be wrong. When /etc/debian_version
> contains "potato/sid", the codename is either potato xor sid, and only apt-
> cache can discriminate a testing host from a sid host. Therefore, in such a
> situation, the correct answer is actually "I can't tell", aka "n/a".

From testing, it reads the information from these files:

  - /var/lib/apt/lists/deb.debian.org_debian_dists_bullseye_InRelease
  - /var/lib/apt/lists/deb.debian.org_debian_dists_bullseye_main_binary-amd64_Packages.lz4

Removing these files (`rm -rf /var/lib/apt/lists/*`) causes 'lsb_release -sc`
to print 'n/a'. While we could use '/etc/debian_version' as a fallback for our
own scripts (stripping everything after '/' (e.g. bullseye/sid -> bullseye),
dpkg-buildpackage will still depend on this information to be present, and
if not present, renames packages to use 'n/a' in their path:

    dpkg-buildpackage: info: full upload; Debian-native package (full source is included)
    renamed '../containerd.io-dbgsym_0.20210219.014044~e58be59-1_amd64.deb' -> '/build/debian/n/a/amd64/containerd.io-dbgsym_0.20210219.014044~e58be59-1_amd64.deb'
    renamed '../containerd.io_0.20210219.014044~e58be59-1_amd64.deb' -> '/build/debian/n/a/amd64/containerd.io_0.20210219.014044~e58be59-1_amd64.deb'

Given that we don't need the final image (as we only use it as a build environment
and copy the artifacts out), keeping some of the cache files should not be a problem.

### DIFF
--- a/dockerfiles/deb.dockerfile
+++ b/dockerfiles/deb.dockerfile
@@ -56,10 +56,41 @@ COPY --from=go-md2man /go/bin/go-md2man /go/bin/go-md2man
 COPY debian/ debian/
 # NOTE: not using a cache-mount for apt, to prevent issues when building multiple
 #       distros on the same machine / build-cache
+#
+# NOTE: DO NOT REMOVE '/var/lib/apt/lists/', to allow building for Debian unstable.
+#
+# Debian "unstable" releases use apt caching information to get the codename
+# see discussion on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651:
+#
+# > That's all to say that this bug is (to my belief) actually expected behaviour;
+# > and fixing it through forcing the codename to be interpreted as "stretch" when
+# > apt-cache information is unavailable would be wrong. When /etc/debian_version
+# > contains "potato/sid", the codename is either potato xor sid, and only apt-
+# > cache can discriminate a testing host from a sid host. Therefore, in such a
+# > situation, the correct answer is actually "I can't tell", aka "n/a".
+#
+# From testing on https://github.com/docker/containerd-packaging/pull/213#issuecomment-782172567,
+# it reads the information from these files:
+#
+#   - /var/lib/apt/lists/deb.debian.org_debian_dists_bullseye_InRelease
+#   - /var/lib/apt/lists/deb.debian.org_debian_dists_bullseye_main_binary-amd64_Packages.lz4
+#
+# Removing these files (`rm -rf /var/lib/apt/lists/*`) causes 'lsb_release -sc`
+# to print 'n/a'. While we could use '/etc/debian_version' as a fallback for our
+# own scripts (stripping everything after '/' (e.g. bullseye/sid -> bullseye),
+# dpkg-buildpackage will still depend on this information to be present, and
+# if not present, renames packages to use 'n/a' in their path:
+#
+#    dpkg-buildpackage: info: full upload; Debian-native package (full source is included)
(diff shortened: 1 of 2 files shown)


dockerfiles/deb.dockerfile
```diff
@@ -56,10 +56,41 @@ COPY --from=go-md2man /go/bin/go-md2man /go/bin/go-md2man
...
+#
+#    dpkg-buildpackage: info: full upload; Debian-native package (full source is included)
+#    renamed '../containerd.io-dbgsym_0.20210219.014044~e58be59-1_amd64.deb' -> '/build/debian/n/a/amd64/containerd.io-dbgsym_0.20210219.014044~e58be59-1_amd64.deb'
+#    renamed '../containerd.io_0.20210219.014044~e58be59-1_amd64.deb' -> '/build/debian/n/a/amd64/containerd.io_0.20210219.014044~e58be59-1_amd64.deb'
+#
+# Given that we don't need the final image (as we only use it as a build environment
+# and copy the artifacts out), keeping some of the cache files should not be a problem.
 RUN apt-get update -q \
  && mk-build-deps -t "apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends -y" -i debian/control \
  && apt-get clean \
```
tianon, line 92: :eyes: this is a no-op for sure:

https://github.com/debuerreotype/debuerreotype/blob/5cf7949ecf1cec1afece267688bda64cd34a6817/scripts/debuerreotype-minimizing-config#L85-L109

dockerfiles/deb.dockerfile
```diff
@@ -56,10 +56,41 @@ COPY --from=go-md2man /go/bin/go-md2man /go/bin/go-md2man
...
- && rm -rf /var/cache/apt /var/lib/apt/lists/*
+ && rm -rf /var/cache/apt
```
tianon, line 93: ... which means this is a no-op too (and redundant with `apt-get clean` in a normal install anyhow)

tianon (approved):
I guess https://github.com/docker/containerd-packaging/pull/220#pullrequestreview-594448879 means I give this a transitive LGTM :sweat_smile:

scripts/build-deb
```diff
@@ -55,6 +55,14 @@ DEST_DIR="/build/${DIST_ID}/${DIST_VERSION}/${ARCH}/"
 mkdir -p "${DEST_DIR}"
 mv -v ../*.deb "${DEST_DIR}"
 
+if [ "${DIST_VERSION}" = 'n/a' ]; then
+	# Give a more useful error message if we failed to get the "codename",
+	# which can be the case on Debian "unstable" releases. For more background:
+	# https://github.com/docker/containerd-packaging/pull/213#issuecomment-782172567
+	echo "Failed to get release codename";
+	exit 1;
```
tianon, line 63: Oh wait, these don't need semicolons :innocent:

thaJeztah:
I'll take your LGTM (CI is 💚)

scripts/build-deb
thaJeztah: oh! cr*p. Fixing

tianon (approved):
re-LGTM :+1:

tiborvass (approved):

PR #222, opened 2021-03-04
Jenkins parameters belong to Jenkinsfile

tiborvass:
For some unknown reasons there are Jenkins parameters that are neither defined in Jenkins nor in Jenkinsfile.

ARCHIVE and CONTAINERD_REF were taken from the output of:
https://ci-next.docker.com/teams-core/job/containerd-packaging/job/master/configure

while RUNC_REF was added so we can build a version different from the vendored one in containerd.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,17 @@
 #!groovy
 
+properties(
+    [
+        parameters(
+            [
+                booleanParam(name: 'ARCHIVE', defaultValue: false, description: 'Archive the build artifacts by pushing to an S3 bucket.'),
+                string(name: 'CONTAINERD_REF', defaultValue: 'master', description: 'Git ref of containerd repo to build.'),
+                string(name: 'RUNC_REF', defaultValue: '', description: 'Git ref of runc repo to build.'),
+            ]
+        )
+    ]
+)
+
 // List of packages to build. Note that this list is overridden in the packaging
 // repository, where additional variants may be added for enterprise.
 //


StefanScherer (approved):
Makes sense to put this into code! 🎉 
LGTM

Jenkinsfile
```diff
@@ -1,5 +1,17 @@
 #!groovy
 
+properties(
+    [
+        parameters(
+            [
+                booleanParam(name: 'ARCHIVE', defaultValue: false, description: 'Archive the build artifacts by pushing to an S3 bucket.'),
+                string(name: 'CONTAINERD_REF', defaultValue: 'master', description: 'Git ref of containerd repo to build.'),
+                string(name: 'RUNC_REF', defaultValue: '', description: 'Git ref of runc repo to build.'),
```
thaJeztah, line 9: Will this cause the default "automatic selection from upstream" to be skipped? https://github.com/docker/containerd-packaging/blob/484fe7cdeaeb2711cb0a823bff4d5ee51009f0ff/common/common.mk#L27

Jenkinsfile
```diff
@@ -1,5 +1,17 @@
 #!groovy
 
+properties(
+    [
+        parameters(
+            [
+                booleanParam(name: 'ARCHIVE', defaultValue: false, description: 'Archive the build artifacts by pushing to an S3 bucket.'),
```
thaJeztah, line 7: Is this used anywhere in this repository? I know we have `CREATE_ARCHIVE`, but that's not related to pushing to s3; https://github.com/docker/containerd-packaging/blob/8c249eac8b8c9098ae42d1a6d828c1d8912cb702/Jenkinsfile#L52


(Couldn't find a reference to this parameter in our internal release repo either)

thaJeztah (commented):
> For some unknown reasons there are Jenkins parameters that are neither defined in Jenkins nor in Jenkinsfile.

@tiborvass I think these were not in this Jenkinsfile because the Jenkinsfile in this CI is only used to build "tip". 

The Jenkinsfile in our release pipeline has the options to pass what versions to build from when we do a package release; https://github.com/docker/release-packaging/blob/containerd/Jenkinsfile#L96-L104

(copying the snippet here, as it's in a private repo);

```groovy
    parameters {
        string(name: 'CONTAINERD_REF',    defaultValue: 'v1.4.2',                                     description: 'containerd reference to build (usually a tagged version)')
        string(name: 'CONTAINERD_REMOTE', defaultValue: 'git@github.com:containerd/containerd',       description: 'repo containing containerd source code')
        string(name: 'PACKAGING_REF',     defaultValue: 'master',                                     description: 'release-packaging reference to build')
        string(name: 'PACKAGING_REMOTE',  defaultValue: 'git@github.com:docker/containerd-packaging', description: 'repo containing code to package containerd')
        string(name: 'GO_VERSION',        defaultValue: '1.13.15',                                    description: 'Go version to use to build')
    }
```

Looking at that, I think we need to update that Jenkinsfile to allow overriding `RUNC_REF` (which, so far, we didn't do)

PR #223, opened 2021-03-08
Add changelog for containerd v1.4.4-1

fuweid:
Signed-off-by: Wei Fu <fuweid89@gmail.com>

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+containerd.io (1.4.4-1) release; urgency=high
+
+  * Update to containerd 1.4.4 to address CVE-2021-21334.
+
+ -- Wei Fu <fuweid89@gmail.com> Mon, 08 Mar 2021 14:14:44 +0000
+
 containerd.io (1.4.3-2) release; urgency=medium
 
   * Update runc to v1.0.0-rc93
--- a/rpm/containerd.spec
+++ b/rpm/containerd.spec
@@ -157,6 +157,9 @@ done
 
 
 %changelog
+* Mon Mar 08 2021 Wei Fu <fuweid89@gmail.com> - 1.4.4-3.1
+- Update to containerd 1.4.4 to address CVE-2021-21334.
+
 * Wed Mar 03 2021 Tibor Vass <tibor@docker.com> - 1.4.3-3.2
 - Update runc to v1.0.0-rc93
 


fuweid:
ping @thaJeztah @tonistiigi @tiborvass ptal thanks~

debian/changelog
```diff
@@ -1,3 +1,9 @@
+containerd.io (1.4.4-1) release; urgency=medium
```
thaJeztah, line 1: Perhaps good to change this to `urgency=high` (I know we did that on previous security releases of upstream containerd)

thaJeztah (changes requested):
Thanks! Had this one on my list for today, so this saves me some work 😅; left one comment

debian/changelog
fuweid: Make senses to me. Updated~ Thanks

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
Hmm.. I notice that https://github.com/docker/containerd-packaging/pull/221 was also merged, but looks like packages were not (yet) pushed to download.docker.com; I'll check with @tiborvass if that was intentional (and if we'll skip that version, or push both);

```bash
$ apt-cache madison containerd.io | head -n4
containerd.io |    1.4.3-1 | https://download.docker.com/linux/ubuntu bionic/stable amd64 Packages
containerd.io |    1.3.9-1 | https://download.docker.com/linux/ubuntu bionic/stable amd64 Packages
containerd.io |    1.3.7-1 | https://download.docker.com/linux/ubuntu bionic/stable amd64 Packages
containerd.io |   1.2.13-2 | https://download.docker.com/linux/ubuntu bionic/stable amd64 Packages
```

I suspect that relates to https://github.com/docker/containerd-packaging/pull/222#pullrequestreview-606056893

fuweid:
> https://github.com/docker/containerd-packaging/pull/221

just FYI, containerd doesn't mention runc-rc93 in vendor.conf 😂

thaJeztah:
> just FYI, containerd doesn't mention runc-rc93 in vendor.conf 😂

Yes, so there's two things currently; current code in this repo was not yet adjusted for the switch to `go.mod` in upstream (had that one on my list; just pushed a fix, but looks like I need to modify somethings; https://github.com/docker/containerd-packaging/pull/224)

Reason we were already moving to runc-rc93 (slightly ahead of upstream containerd) is that there were too many issues related to seccomp when using images with recent ubuntu versions as base image. Runc rc93 has this patch https://github.com/opencontainers/runc/pull/2750, which helps for those situations; we know some cloud providers also started to install runc rc93 (and some distros started to patch their packages to update to that version as well);

tiborvass (approved):

PR #224, opened 2021-03-08
Fix detection of default runc version due upstream to switch to go.mod

thaJeztah:
The code that was used to automatically select the default version of runc to use for containerd was still assuming the containerd code to have a vendor.conf.

Containerd "master", and v1.5 have switched to go modules, so should now look in the go.mod file to get the recommended / default version.

Also change the order in which we check out source-code, so that the containerd source-code is checked out before we determine the runc version; this allows us to always use the local source of containerd, without having to use curl to get the version from GitHub.

Finally, a small modification was made for the "make validate" target, which now uses a generic "alpine" Golang image, so that validation can be done before the containerd source code was checked out (which was used to detect the Go version to use).

Before this patch:

    $ make docker.io/library/ubuntu:focal
    ...
    curl: (22) The requested URL returned error: 404 Not Found
    --------------------------------------------------------------------
    Building packages on docker.io/library/ubuntu:focal

    containerd   : HEAD (commit: a72fe7d)
    runc         :  (commit: 59ad417)
    architecture : x86_64
    build image  : docker.io/library/ubuntu:focal
    golang image : docker.io/library/golang:1.15.8-buster

With this patch:

    $ make docker.io/library/ubuntu:focal
    ...
    Building packages on docker.io/library/ubuntu:focal

    containerd   : HEAD (commit: a72fe7d)
    runc         : v1.0.0-rc93 (commit: 59ad417)
    architecture : x86_64
    build image  : docker.io/library/ubuntu:focal
    golang image : docker.io/library/golang:1.15.8-buster

### DIFF
--- a/common/common.mk
+++ b/common/common.mk
@@ -19,17 +19,8 @@ RUNC_REMOTE       ?=https://github.com/opencontainers/runc.git
 REF?=HEAD
 
 # Select the default version of Golang and runc based on the containerd source.
-# The runc version commit/tag is defined in vendor.conf. Code below is based on
-# the code used in the containerd repository:
-# https://github.com/containerd/containerd/blob/499fbb0337c9138b5360117e0b25a7a1428f9667/script/setup/install-runc#L24
-ifdef CONTAINERD_DIR
-GOVERSION?=$(shell grep "ARG GOLANG_VERSION" $(CONTAINERD_DIR)/contrib/Dockerfile.test | awk -F'=' '{print $$2}')
-RUNC_REF?=$(shell grep "opencontainers/runc" "$(CONTAINERD_DIR)/vendor.conf" | awk '{print $$2}')
-else
-# TODO adjust GOVERSION and RUNC_REF macro to take CONTAINERD_REMOTE into account
-GOVERSION?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/contrib/Dockerfile.test" | grep "ARG GOLANG_VERSION" | awk -F'=' '{print $$2}')
-RUNC_REF?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/vendor.conf" | grep "opencontainers/runc" | awk '{print $$2}')
-endif
+GOVERSION?=$(shell grep "ARG GOLANG_VERSION" src/github.com/containerd/containerd/contrib/Dockerfile.test | awk -F'=' '{print $$2}')
+RUNC_REF?=$(shell scripts/determine-runc-version)
 
 GOLANG_IMAGE=golang:$(GOVERSION)
 ifeq ($(OS),Windows_NT)
--- a/scripts/determine-runc-version
+++ b/scripts/determine-runc-version
@@ -1,0 +1,66 @@
+#!/usr/bin/env sh
+
+#   Copyright 2018-2020 Docker Inc.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+
+#       http://www.apache.org/licenses/LICENSE-2.0
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+# Select the default version of runc based on the containerd source if no
+# RUNC_REF is set manually. For containerd > 1.5.0-beta.4, and containerd > 1.4.4,
+# the runc version commit/tag is defined script/setup/runc-version. For older
+# versions, use go.mod or vendor.conf.
+#
+# See the install-runc script in the containerd repository:
+# https://github.com/containerd/containerd/blob/v1.5.0-beta.4/script/setup/install-runc#L24-L27
+# https://github.com/containerd/containerd/blob/v1.5.0-beta.3/script/setup/install-runc#L24
+# https://github.com/containerd/containerd/blob/v1.4.0/script/setup/install-runc#L24
+runc_version() {
+	if [ -n "${RUNC_REF}" ]; then
+		# just a safe-guard if this script is called when RUNC_REF was already set.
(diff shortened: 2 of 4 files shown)


thaJeztah:
Hmm... looks like `make validate` now fails; let me fix that

```
+ make validate
grep: src/github.com/containerd/containerd/contrib/Dockerfile.test: No such file or directory
docker run --rm -v /home/ubuntu/workspace/containerd-packaging_PR-224:/work -w /work docker.io/library/golang:-buster bash -c 'go get -u github.com/kunalkushwaha/ltag && ./scripts/validate/fileheader'
docker: invalid reference format.
See 'docker run --help'.
Makefile:113: recipe for target 'validate' failed
make: *** [validate] Error 125
script returned exit code 2
```

thaJeztah:
Fixed the `make validate` step to no longer depend on the containerd source code, and instead just use `golang:alpine`

thaJeztah:
Looks good now; https://ci-next.docker.com/teams-core/blue/rest/organizations/jenkins/pipelines/containerd-packaging/branches/PR-224/runs/2/nodes/88/steps/605/log/?start=0

```
+ git -C src/github.com/opencontainers/runc checkout -q FETCH_HEAD
# upstream systemd unit uses /usr/local/bin, whereas our packages use /usr/bin
sed 's#/usr/local/bin/containerd#/usr/bin/containerd#g' src/github.com/containerd/containerd/containerd.service > common/containerd.service
--------------------------------------------------------------------
Building packages on docker.io/library/ubuntu:focal

containerd   : HEAD (commit: a72fe7d)
runc         : v1.0.0-rc93 (commit: 59ad417)
architecture : aarch64
build image  : docker.io/library/ubuntu:focal
golang image : docker.io/library/golang:1.15.8-buster
--------------------------------------------------------------------
focal: Pulling from library/ubuntu
```

thaJeztah:
@silvin-lubecki @fuweid @StefanScherer @tianon @tiborvass ptal


note that we'll have to update this script once https://github.com/containerd/containerd/pull/5036 has been accepted upstreeam

scripts/determine-runc-version
```diff
@@ -0,0 +1,45 @@
...
+# https://github.com/containerd/containerd/blob/v1.5.0-beta.3/script/setup/install-runc#L24
+# https://github.com/containerd/containerd/blob/v1.4.0/script/setup/install-runc#L24
+runc_version() {
+	if [ -n "${RUNC_REF}" ]; then
+		echo "${RUNC_REF}"
+		return
+	fi
+
+	# shellcheck disable=SC2164
+	repo_abspath="$(cd -- "$(dirname "$0")/.." >/dev/null 2>&1 ; pwd -P)"
```
thaJeztah, line 31: don't know if there's a cleaner way; this variant seems to be portable (as in; it works on macOS as well, which doesn't support `readlink -f`)

common/common.mk
```diff
@@ -19,17 +19,8 @@ RUNC_REMOTE       ?=https://github.com/opencontainers/runc.git
...
-# https://github.com/containerd/containerd/blob/499fbb0337c9138b5360117e0b25a7a1428f9667/script/setup/install-runc#L24
-ifdef CONTAINERD_DIR
-GOVERSION?=$(shell grep "ARG GOLANG_VERSION" $(CONTAINERD_DIR)/contrib/Dockerfile.test | awk -F'=' '{print $$2}')
-RUNC_REF?=$(shell grep "opencontainers/runc" "$(CONTAINERD_DIR)/vendor.conf" | awk '{print $$2}')
-else
-# TODO adjust GOVERSION and RUNC_REF macro to take CONTAINERD_REMOTE into account
-GOVERSION?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/contrib/Dockerfile.test" | grep "ARG GOLANG_VERSION" | awk -F'=' '{print $$2}')
-RUNC_REF?=$(shell curl -fsSL "https://raw.githubusercontent.com/containerd/containerd/$(REF)/vendor.conf" | grep "opencontainers/runc" | awk '{print $$2}')
-endif
+GOVERSION?=$(shell grep "ARG GOLANG_VERSION" src/github.com/containerd/containerd/contrib/Dockerfile.test | awk -F'=' '{print $$2}')
```
fuweid, line 22: After checked the release/1.4.4, the `contrib/Dockerfile.test` isn't updated to go1.15.8. Is it possible to use `.github/workflows/release.yml`?

scripts/determine-runc-version
fuweid: Just curiosity about `--`.  look likes `repo_abspath="$(cd "$(dirname "$0")/.." >/dev/null 2>&1 ; pwd -P)"` can work.

thaJeztah: Good point; TBH I was testing some variations that didn't work on macOS, and found this one, which did work, but was also wondering why the `--` was there, and forgot to test without; let me try

common/common.mk
thaJeztah: Hmm.. I think we should fix that; if the Dockerfile.test is out of sync, that's a mistake

thaJeztah: Actually; it looks up-to-date; https://github.com/containerd/containerd/blob/v1.4.4/contrib/Dockerfile.test#L9

```dockerfile
ARG GOLANG_VERSION=1.15.8
```

That said; I think this should be treated as a default (to make things work "out of the box"), but probably the recommendation should be to pass these variables when building 🤔

scripts/determine-runc-version
thaJeztah: Looks like it's the preferred way to do this (to be most portable); there's some discussion in https://go-review.googlesource.com/c/sys/+/210157/, which at some point suggests adding the `--` instead of skipping it.

silvin-lubecki (dismissed):
Looks good 👍

common/common.mk
fuweid: Sorry my mistake. I used the makefile to build and the script uses 1.15.5 goversion...

fuweid (dismissed):
LGTM(nb)

thaJeztah:
Updated to take the changes from upstream https://github.com/containerd/containerd/pull/5036 into account

silvin-lubecki (approved):
LGTM

tiborvass (approved):

PR #233, opened 2021-05-19
Revert "remove default "containerd.toml" config file"

thaJeztah:
This reverts commit 37e406ca6302b373b5840b794cf551110ad4d737.

Relates to https://github.com/docker/containerd-packaging/pull/215#discussion_r632804641

> removing this does have some impact on rpm users who are currently overwriting
> the rpm supplied config.toml with their own. This being removed above, causes
> the user supplied override config.toml file also to be deleted, when they upgrade
> to a rpm version with this change. rpm moves the file to.rpmsave/.rpmnew after
> upgrading to the new version with the above change.

Reverting this change (at least temporarily until we figure out an alternative)
is the safest option for now.

### DIFF
--- a/common/containerd.toml
+++ b/common/containerd.toml
@@ -1,0 +1,31 @@
+#   Copyright 2018-2020 Docker Inc.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+
+#       http://www.apache.org/licenses/LICENSE-2.0
+
+#   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


thaJeztah:
@ravilr @tianon @silvin-lubecki ptal

tianon (approved):
"yay"

LGTM

(is it worth reverting just the rpm bits?  no strong preference from me either way)

thaJeztah:
Yes, I was considering just to revert only the RPM changes, but it's worth looking into "other options" (if possible); if not, then we could also consider making this change for the v1.5.x release

fredericdalleau (approved):
LGTM

PR #235, opened 2021-05-21
Fix RUNC_REF not being evaluated correctly

thaJeztah:
When testing the script locally from a clean state, I noticed that the RUNC_REF
variable was not working as expected, and failed to pick the runc version that
is set in containerd.

This patch fixes the runc version not being detected correctly.

While Make "promises" to expand variables "when used", it actually performs
this earlier than that, so when combining, it evaluates `$(SOME_VAR)` *before*
the target is run....

From the docs: https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors

> references are expanded whenever this variable is substituted (in the course
> of expanding some other string). When this happens, it is called recursive expansion.

However, taking the following Makefile:

```make
SOME_VAR=$(shell cat hello)

.PHONY: clean
clean:
    @rm hello

.PHONY: together
together:
    @echo this is hello > hello
    @cat hello
    @echo SOME_VAR is \""$(SOME_VAR)"\"

.PHONY: separate
separate: one two

.PHONY: one
one:
    @echo this is hello > hello
    @cat hello

.PHONE: two
two:
    @echo SOME_VAR is \""$(SOME_VAR)"\"
```

When combined:

```bash
$ make clean
$ make together
cat: hello: No such file or directory
this is hello
SOME_VAR is ""
```

As shown, `$(SOME_VAR)` is evaluated *before* `together` is run, and produces
an error, because "hello" is not yet created.

When doing creating the file and using the file separate, things work as
expected, and `$(SOME_VAR)` is able to read the file:

```bash
$ make clean
$ make separate
this is hello
SOME_VAR is "this is hello"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,15 @@ docker.io/%:
 	$(MAKE) BUILD_IMAGE="$@" build
 
 .PHONY: checkout
-checkout: src
+checkout: checkout-containerd checkout-runc
+
+.PHONY: checkout-containerd
+checkout-containerd: src
 	./scripts/checkout.sh src/github.com/containerd/containerd "$(REF)"
+
+# this must be a separate target, otherwise "RUNC_REF" is not evaluated correctly
+.PHONY: checkout-runc
+checkout-runc: checkout-containerd
 	./scripts/checkout.sh src/github.com/opencontainers/runc "$(RUNC_REF)"
 
 .PHONY: build
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,6 @@ BUILD_BASE=$(shell ./scripts/determine-base $(BUILD_IMAGE))
 # For example, use "make PROGRESS=plain ..." to show build progress in plain test
 PROGRESS=auto
 TARGET=packages
-CONTAINERD_COMMIT=$(shell git -C "src/github.com/containerd/containerd" log -1 --pretty='%h')
-RUNC_COMMIT=$(shell git -C "src/github.com/opencontainers/runc" log -1 --pretty='%h')
 
 all: build
 
@@ -72,16 +70,9 @@ docker.io/%:
 	$(MAKE) BUILD_IMAGE="$@" build
 
 .PHONY: checkout
-checkout: checkout-containerd checkout-runc
-
-.PHONY: checkout-containerd
-checkout-containerd: src
+checkout: src
 	./scripts/checkout.sh src/github.com/containerd/containerd "$(REF)"
-
-# this must be a separate target, otherwise "RUNC_REF" is not evaluated correctly
-.PHONY: checkout-runc
-checkout-runc: checkout-containerd
-	./scripts/checkout.sh src/github.com/opencontainers/runc "$(RUNC_REF)"
+	./scripts/checkout.sh src/github.com/opencontainers/runc "$$(./scripts/determine-runc-version)"
 
 .PHONY: build
 build: checkout common/containerd.service
@@ -89,8 +80,8 @@ build:
 	@echo "--------------------------------------------------------------------"
 	@echo "Building $(TARGET) on $(BUILD_IMAGE)"
 	@echo ""
-	@echo "containerd   : $(REF) (commit: $(CONTAINERD_COMMIT))"
-	@echo "runc         : $(RUNC_REF) (commit: $(RUNC_COMMIT))"
+	@echo "containerd   : $(REF) (commit: $(shell git -C "src/github.com/containerd/containerd" log -1 --pretty='%h'))"
+	@echo "runc         : $$(./scripts/determine-runc-version) (commit: $$(git -C "src/github.com/opencontainers/runc" log -1 --pretty='%h'))"
 	@echo "architecture : $(shell uname -m)"
 	@echo "build image  : $(BUILD_IMAGE)"
 	@echo "golang image : $(GOLANG_IMAGE)"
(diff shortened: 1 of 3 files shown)


thaJeztah:
Looking good now;

```
./scripts/checkout.sh src/github.com/opencontainers/runc "v1.0.0-rc95"
+ SRC=src/github.com/opencontainers/runc
+ REF=v1.0.0-rc95
+ REF_FETCH=v1.0.0-rc95
+ awk {print $2}
+ git -C src/github.com/opencontainers/runc ls-remote --refs --heads --tags origin v1.0.0-rc95
+ REF=refs/tags/v1.0.0-rc95
+ [ -n refs/tags/v1.0.0-rc95 ]
+ REF_FETCH=refs/tags/v1.0.0-rc95:refs/tags/v1.0.0-rc95
+ git -C src/github.com/opencontainers/runc fetch --update-head-ok --depth 1 origin refs/tags/v1.0.0-rc95:refs/tags/v1.0.0-rc95
From https://github.com/opencontainers/runc
 * [new tag]         v1.0.0-rc95 -> v1.0.0-rc95
+ git -C src/github.com/opencontainers/runc checkout -q refs/tags/v1.0.0-rc95
# upstream systemd unit uses /usr/local/bin, whereas our packages use /usr/bin
sed 's#/usr/local/bin/containerd#/usr/bin/containerd#g' src/github.com/containerd/containerd/containerd.service > common/containerd.service
INFO: detected runc version from script/setup/runc-version
--------------------------------------------------------------------
Building packages on docker.io/library/centos:8

containerd   : HEAD (commit: a0efc54)
runc         : v1.0.0-rc95 (commit: b9ee9c6)
architecture : x86_64
build image  : docker.io/library/centos:8
golang image : docker.io/library/golang:1.16.4-buster
--------------------------------------------------------------------
```

fredericdalleau:
It looks like the patch relies on the fact that the RUNC_REF is evaluated just before the recipe is built and on the fact that recipes for different make targets are not built at the same time. Is there any reference to this behavior in the Makefile documentation?

Have you considered using `scripts/determine-runc-version` between backticks instead so that it gets evaluated by the shell during the recipe instead of make ?

thaJeztah:
> Have you considered using scripts/determine-runc-version between backticks instead so that it gets evaluated by the shell during the recipe instead of make ?

Yes, I was considering that; the idea of taking the current approach would be that it's evaluated / set when checking out the source code for runc. The variable can either be passed as argument (to override the version), and otherwise would use the version specified in the containerd source-code as default.

I'm not entirely sure if that makes it more readable though (I just pushed a commit with something like that)

Makefile
```diff
@@ -72,25 +70,18 @@ docker.io/%:
...
-.PHONY: checkout-containerd
-checkout-containerd: src
+checkout: src
 	./scripts/checkout.sh src/github.com/containerd/containerd "$(REF)"
-
-# this must be a separate target, otherwise "RUNC_REF" is not evaluated correctly
-.PHONY: checkout-runc
-checkout-runc: checkout-containerd
-	./scripts/checkout.sh src/github.com/opencontainers/runc "$(RUNC_REF)"
+	./scripts/checkout.sh src/github.com/opencontainers/runc "$(shell ./scripts/determine-runc-version)"
```
fredericdalleau, line 75: I was thinking something like this :
```
./scripts/checkout.sh src/github.com/opencontainers/runc `./scripts/determine-runc-version`
```

thaJeztah: Yeah, trying to avoid the backticks as they aren't posix (and IIRC officially deprecated? (at least "discouraged")), but using a plain `$()` doesn't work in Make without using `$(shell ...)` 😞

fredericdalleau: You might wanna try `$$(shell command)`

fredericdalleau: And it turns out it correctly uses RUNC_REF from `make RUNC_REF=1.2.3.4`. Btw, I'd suggest echoing the selected RUNC_REF in `determine-runc-version`.

thaJeztah: > You might wanna try `$$(./scripts/determine-runc-version)`

🤦 you're right. for some reason I was thinking that was how to handle it in Jenkinsfile, not in Makefile


> Btw, I'd suggest echoing the selected RUNC_REF in `determine-runc-version`.

Not sure I understand what you mean; it should already print the version it found, and print an informational message on `stderr` but perhaps that won't be visible due to the sub shell; https://github.com/docker/containerd-packaging/blob/6a2e254c9b238a482c5bdebe8e5d99386522515a/scripts/determine-runc-version#L26-L43

thaJeztah: Oh! You mean print the version as part of the `INFO:` messages? Yes, that makes sense

fredericdalleau (approved):
LGTM

Makefile
```diff
@@ -72,29 +70,23 @@ docker.io/%:
...
 	@echo ""
-	@echo "containerd   : $(REF) (commit: $(CONTAINERD_COMMIT))"
-	@echo "runc         : $(RUNC_REF) (commit: $(RUNC_COMMIT))"
+	@echo "containerd   : $(REF) (commit: $(shell git -C "src/github.com/containerd/containerd" log -1 --pretty='%h'))"
+	@echo "runc         : $$(./scripts/determine-runc-version) (commit: $$(git -C "src/github.com/opencontainers/runc" log -1 --pretty='%h'))"
 	@echo "architecture : $(shell uname -m)"
 	@echo "build image  : $(BUILD_IMAGE)"
 	@echo "golang image : $(GOLANG_IMAGE)"
 	@echo "--------------------------------------------------------------------"
+	exit 1;
```
fredericdalleau, line 89: Oops

fredericdalleau (changes requested):
You need to remove the exit I think

Makefile
thaJeztah: ah, lol, yes, was debugging

thaJeztah: fixed 👍

fredericdalleau (approved):
LGTM

PR #238, opened 2021-07-03
deb: add override_dh_strip to disable stripping binaries

thaJeztah:
We had this override in place for the docker engine packaging (https://github.com/docker/docker-ce-packaging/blob/4cd3a7eca2f7887b15e98c2e67be1880f337d291/deb/common/rules#L42-L43, and in upstream moby (before packaging was moved); https://github.com/moby/moby/blob/v1.10.0/hack/make/.build-deb/rules#L19-L21), and I noticed it was missing here. This looks still relevant though, so adding here as well.

Comment was taken from the https://github.com/git-lfs/git-lfs repository, which had a link with more information :)

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -49,6 +49,10 @@ man: ## Create containerd man pages
 
 override_dh_auto_build: binaries bin/runc man
 
+override_dh_strip:
+	# strip disabled as golang upstream doesn't support it and it makes go
+	# crash. See https://launchpad.net/bugs/1200255.
+
 override_dh_systemd_start:
 	dh_systemd_start --restart-after-upgrade
 	sed -i 's/_dh_action=try-restart/_dh_action=restart/g' ./debian/containerd.io.postinst.debhelper


thaJeztah:
@tianon @silvin-lubecki ptal

tianon:
I think there was a discussion on this in runc recently, right?  I think it was very relevant in Go 1.6 but may not actually be relevant anymore?

thaJeztah:
Oh, I might've missed that discussion. I was working on the other repository, and noticed that this one didn't have it. Then I also recalled some issues with armhf binaries at some point, and wondered if those could be related 😅

tianon:
Yep, see https://github.com/opencontainers/runc/pull/3054#issuecomment-871887158:

> GNU `strip` used to break stack backtraces in Go programs, while `-w` would reduce the size of the binary without that problem (in fact there are some go-nuts threads from a long time ago where folks were told to not use GNU strip on Go binaries).

So, I think this should be fine to let `dh_strip` do its thing?

The issues with ARM binaries was with the static binaries (not the `.deb` packages); https://github.com/docker-library/docker/issues/260

thaJeztah:
Ah, interesting.

Trying to find it that issue has any references (other than "Now, it seems that this is no longer the case" 😅). If true, I guess we can remove it from https://github.com/docker/docker-ce-packaging/blob/4cd3a7eca2f7887b15e98c2e67be1880f337d291/deb/common/rules#L42-L43 as well 🤔

tianon:
A bit hard to search for, but I found http://dominik.honnef.co/posts/2016/10/go-and-strip/ which is full of a bunch of useful references that say "yes, stripping is fine (maybe unless you use `gccgo`)", especially https://groups.google.com/g/golang-dev/c/ABppMOjYP6w/m/2RXbcSQPAwAJ.

PR #240, opened 2021-07-12
RPM : Update libseccomp dependency name for SLES/openSUSE distros

anujajakhade:
Added check for suse_version to install dependencies.

### DIFF
--- a/rpm/containerd.spec
+++ b/rpm/containerd.spec
@@ -59,6 +59,10 @@ Source3: runc
 Requires: container-selinux >= 2:2.74
 %endif
 Requires: libseccomp
+%else
+# SUSE flavors do not have container-selinux,
+# and libseccomp is named libseccomp2
+Requires: libseccomp2
 %endif
 BuildRequires: make
 BuildRequires: gcc


rpm/containerd.spec
```diff
@@ -58,8 +58,12 @@ Source3: runc
 %if "%{?dist}" != ".amzn2"
 Requires: container-selinux >= 2:2.74
 %endif
+%if 0%{?suse_version}
+Requires: libseccomp2
+%else
```
thaJeztah, line 63: Do we need a similar change in docker-ce-packaging? https://github.com/docker/docker-ce-packaging/blob/5d9ad8ccf9251201a8b436bd09435536e3cfd309/rpm/SPECS/docker-ce.spec#L22

Also wondering is the package name for `libseccomp-devel` also different on SUSE? (e.g. `libseccomp2-devel`);

https://github.com/docker/docker-ce-packaging/blob/5d9ad8ccf9251201a8b436bd09435536e3cfd309/rpm/SPECS/docker-ce.spec#L39

anujajakhade: @thaJeztah Yes, we need to do a similar change in docker-ce-packaging.
I'll be raising a PR for the changes in docker-ce-packaging soon.

`libseccomp2-devel` is not present in SUSE. We need to install libseccomp-devel.

thaJeztah: Actually; I noticed we have the `%if %{undefined suse_version}` a few lines up. I wanted to comment "can we use the same for this one", but then noticed that probably this if/else is not in the right place.

It's unfortunate that rpm SPEC files don't allow indenting, otherwise it was more clear, but this is what it currently shows if I indent the `if/else` blocks;

```
# container-selinux isn't a thing in suse flavors
%if %{undefined suse_version}
    # amazonlinux2 doesn't have container-selinux either
    %if "%{?dist}" != ".amzn2"
        Requires: container-selinux >= 2:2.74
    %endif
    %if 0%{?suse_version}
        Requires: libseccomp2
    %else
        Requires: libseccomp
    %endif
%endif
```


So the `0%{?suse_version}` is _inside_ the `%if %{undefined suse_version}`, so would never be true.


To double check if I'm correct; Inspecting the package looks to conform that; the opensuse package has neither `container-selinux` nore `libseccomp` as dependency:

```bash
rpm -qp --requires ./opensuse-leap/15/x86_64/containerd.io-0.20210712.042325~a8ef1fc-0.x86_64.rpm
/bin/sh
/bin/sh
/bin/sh
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
rpmlib(TildeInVersions) <= 4.10.0-1
systemd
systemd
systemd
systemd
```

Whereas the CentOS 8 package has both:

```
rpm -qp --provides --requires ./centos/8/x86_64/containerd.io-0.20210712.042325~a8ef1fc-0.el8.x86_64.rpm
/bin/sh
/bin/sh
/bin/sh
container-selinux >= 2:2.74
libseccomp
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
rpmlib(TildeInVersions) <= 4.10.0-1
systemd
systemd
systemd
```




I think this block can be simplified to (with indenting for readability here); perhaps also move the comment within the branches:

```
%if %{undefined suse_version}
    %if "%{?dist}" != ".amzn2"
        # amazonlinux2 doesn't have container-selinux
        Requires: container-selinux >= 2:2.74
    %endif
    Requires: libseccomp
%else
    # SUSE flavors do not have container-selinux,
    # and libseccomp is named libseccomp2
    Requires: libseccomp2
%endif
```

And actual version, without the indenting:

```
%if %{undefined suse_version}
%if "%{?dist}" != ".amzn2"
# amazonlinux2 doesn't have container-selinux
Requires: container-selinux >= 2:2.74
%endif
Requires: libseccomp
%else
# SUSE flavors do not have container-selinux,
# and libseccomp is named libseccomp2
Requires: libseccomp2
%endif
```

thaJeztah (changes requested):
Left a comment inline; I think the change is incorrect, and doesn't add the "required" property to the package (left a suggestion)

anujajakhade:
@thaJeztah I have done the requested changes. 
Can you please check?
Do let me know if any more changes are needed.

thaJeztah:
Whoops, thought I commented on this one: change itself LGTM, but can you amend the commit-message to be more descriptive (in case it needs to be found back?)

Something like:

```
rpm: update libseccomp dependency name for SLES/openSuse distros
```

anujajakhade:
Hi @thaJeztah I have updated the PR name.

thaJeztah:
Can you update the commit message itself as well? (as that's what would show in git blame / git log; having a good title helps in situations where history is viewed outside of github)

anujajakhade:
Hi @thaJeztah sorry i forgot to amend the commit message. I have changed it now. Do let me know if any more changes are needed.

thaJeztah (approved):
LGTM, thanks!

PR #243, opened 2021-07-19
scripts/determine-runc-version: remove fallback for older containerd versions

thaJeztah:
Starting with containerd v1.4.5, and v1.5.0-beta.4, we can now use the
runc-version file:

- https://github.com/containerd/containerd/blob/v1.4.5/script/setup/runc-version
- https://github.com/containerd/containerd/blob/v1.5.0/script/setup/runc-version

There should no longer be a need to build older (patch) releases of the containerd
packages, so we can now remove the fallbacks.

### DIFF
--- a/scripts/determine-runc-version
+++ b/scripts/determine-runc-version
@@ -15,14 +15,10 @@
 #   limitations under the License.
 
 # Select the default version of runc based on the containerd source if no
-# RUNC_REF is set manually. For containerd > 1.5.0-beta.4, and containerd > 1.4.4,
-# the runc version commit/tag is defined script/setup/runc-version. For older
-# versions, use go.mod or vendor.conf.
+# RUNC_REF is set manually.
 #
 # See the install-runc script in the containerd repository:
-# https://github.com/containerd/containerd/blob/v1.5.0-beta.4/script/setup/install-runc#L24-L27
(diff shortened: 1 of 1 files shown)


scripts/determine-runc-version
```diff
@@ -36,33 +32,16 @@ runc_version() {
...
-		runc_ref=$(grep 'opencontainers/runc' "${containerd_src_dir}/vendor.conf" | awk '{print $2}')
-		>&2 echo "INFO: detected runc version (${runc_ref}) from vendor.conf"
-		echo "${runc_ref}"
-		return
 	fi
 
 	# if all else fails
-	>&2 echo "INFO: unable to detect runc version, using HEAD"
+	>&2 echo "WARNING: unable to detect runc version, using HEAD"
 	echo "HEAD"
```
thaJeztah, line 45: As a follow-up, I want to change this into an error (also update `scripts/checkout.sh` to make both `$SRC` and `$REF` required (non-empty) arguments); https://github.com/docker/containerd-packaging/blob/c58c9172edf1b74c548e649a1282b008da7d43ef/scripts/checkout.sh#L17-L20

djs55 (approved):
Makes sense to me.
LGTM

thaJeztah:
👍 bringing this one in

PR #260, opened 2021-11-22
Jenkinsfile: remove s390x steps

thaJeztah:
Packages for s390x are built by IBM, and the RHEL image that we used was outdated/broken, so made CI fail.


fredericdalleau (approved):
LGTM

PR #231, opened 2021-04-30
Use OS native runc rather than bundling on RHEL8

goeranu:
The RPM does provide on runc for most platforms, but not for RHEL8. It
does not work there, since there is native package that provides runc
that lives in a module stream. Yum fails to find or install a non-modular
package when a module provides the same name.

For those with access to RedHat pages, some more information is available
here: https://access.redhat.com/solutions/5588101

A problem with simply omitting the "provides" from the package is that it
will then come in conflict with the native package. It is for example not
possible to install both docker and the native podman in parallel.

This patch changes the contanerd.io package on RHEL8 to require the native
OS version instead of bundling its own copy. With this, parallel installation
of docker and podman is again possible.

Closes docker/packaging#327

PR #270, opened 2022-03-07
Also build for CentOS 9.

Romain-Geissler-1A:
Note: I have removed the btrfs support unconditionally as IMO the old condition "%if 1%{!?el8:1}" was always true, on every distros.

This includes changes inspired from #231.

### DIFF
--- a/rpm/containerd.spec
+++ b/rpm/containerd.spec
@@ -30,8 +30,8 @@ AutoReq: no
 
 Name: containerd.io
 Provides: containerd
-# For some reason on rhel 8 if we "provide" runc then it makes this package unsearchable
-%if 0%{!?el8:1}
+# For some reason on rhel >= 8 if we "provide" runc then it makes this package unsearchable
+%if 0%{?rhel} <= 7
 Provides: runc
 %endif
 
@@ -69,15 +69,6 @@ BuildRequires: gcc
 BuildRequires: systemd
 BuildRequires: libseccomp-devel
 
-# Should only return true if `el8` (rhel8) is NOT defined
-%if 0%{!?el8:1}
-%if 0%{?suse_version}
-BuildRequires: libbtrfs-devel
-%else
-BuildRequires: btrfs-progs-devel
-%endif
-%endif
-
 %{?systemd_requires}
 
 %description
@@ -108,10 +99,7 @@ cd %{_topdir}/BUILD/
 cd %{_topdir}/BUILD
 GO111MODULE=auto make man
 
-BUILDTAGS="seccomp selinux"
-%if 1%{!?el8:1}
-BUILDTAGS="${BUILDTAGS} no_btrfs"
-%endif
+BUILDTAGS="seccomp selinux no_btrfs"
 
 GO111MODULE=auto make -C /go/src/%{import_path} VERSION=%{getenv:VERSION} REVISION=%{getenv:REF} PACKAGE=%{getenv:PACKAGE} BUILDTAGS="${BUILDTAGS}"
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,6 +9,7 @@ def images = [
     [image: "docker.io/library/amazonlinux:2",          arches: ["aarch64"]],
     [image: "docker.io/library/centos:7",               arches: ["amd64", "aarch64"]],
     [image: "quay.io/centos/centos:stream8",            arches: ["amd64", "aarch64"]],
+    [image: "quay.io/centos/centos:stream9",            arches: ["amd64", "aarch64"]],
     [image: "docker.io/library/debian:buster",          arches: ["amd64", "aarch64", "armhf"]], // Debian 10 (EOL: 2024)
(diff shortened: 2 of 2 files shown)


dockerfiles/rpm.dockerfile
```diff
@@ -116,7 +117,8 @@ RUN createrepo /build \
  && rm -rf /build/repodata
 RUN containerd --version
 RUN ctr --version
-RUN runc --version
+# Don't try to run runc on RHEL/CentOS as we don't package it anymore.
+RUN if [ "${BASE}" != "rhel" ] && [ "${BASE}" != "centos" ]; then runc --version; fi
```
thaJeztah, line 121: I'd rather keep this change for a separate PR (sorry haven't been able to spend time to respond on the linked ticket); this definitely needs a wider discussion (and not something we should change overnight).

I'm ok with adding the `stream9` to the list in CI though (note that this Jenkinsfile is used for CI, but the actual matrix for which we built is in our internal release pipeline, so it may not be included (yet) on the next build)

thaJeztah, line 121: I'd rather keep this change for a separate PR (sorry haven't been able to spend time to respond on the linked ticket); this definitely needs a wider discussion (and not something we should change overnight).

I'm ok with adding the `stream9` to the list in CI though (note that this Jenkinsfile is used for CI, but the actual matrix for which we built is in our internal release pipeline, so it may not be included (yet) on the next build)

Romain-Geissler-1A: I have updated this pull request to focus just on CentOS 9. I have split it in two commits, the second commit being now tracked in #272.

Romain-Geissler-1A:
Note: it looks like Ci is failing, but external contributor can't see the logs (Access Denied Romain-Geissler-1A is missing the Global/Read permission), so I don't know why it's failing and can't fix it myself. Locally `make quay.io/centos/centos:stream9` works for me.

Romain-Geissler-1A:
I saw you have release docker 20.10.13 with Ubuntu Jammy this week, would it be possible to also tackle CentOS 9 ?

Romain-Geissler-1A:
Note for maintainers: CI build is failing, but open source contributor can't see the output of Jenkins, so I would need you to tell me what's wrong in order to fix the build.

Romain-Geissler-1A:
Ping (I don't know why the build fails)

Romain-Geissler-1A:
Ping

Romain-Geissler-1A:
Ping

thaJeztah:
I saw failures on https://github.com/docker/docker-ce-packaging/pull/681#issuecomment-1113331534 as well; looks like it's now green, so I'm guessing it was some failures with those package servers (or just general networking issues); let me give Jenkins another kick

rpm/containerd.spec
```diff
@@ -69,15 +69,6 @@ BuildRequires: gcc
...
 BuildRequires: libseccomp-devel
 
-# Should only return true if `el8` (rhel8) is NOT defined
-%if 0%{!?el8:1}
-%if 0%{?suse_version}
-BuildRequires: libbtrfs-devel
-%else
-BuildRequires: btrfs-progs-devel
-%endif
-%endif
```
thaJeztah, line 79: Not sure if this is correct; I think the intent here was to

- include `libbtrfs-devel` on centos / rhel 7 (which still had btrfs?), and Fedora (?)
- include `btrfs-progs-devel` on SUSE flavours (which default to using BTRFS as filesystem)

If I'm reading the current changes correctly, we're now excluding btrfs on all RPM variants (not just RHEL/CentOS > 7

rpm/containerd.spec
```diff
@@ -30,8 +30,8 @@ AutoReq: no
 
 Name: containerd.io
 Provides: containerd
-# For some reason on rhel 8 if we "provide" runc then it makes this package unsearchable
-%if 0%{!?el8:1}
+# For some reason on rhel >= 8 if we "provide" runc then it makes this package unsearchable
+%if 0%{?rhel} <= 7
```
thaJeztah, line 34: This may also be excluding this rule on Fedora, (open)SUSE, Oracle etc (assuming those don't have `rhel` defined), so a larger change than only "exclude on RHEL >= 8

rpm/containerd.spec
Romain-Geissler-1A: Actually the removal of all this related to btrfs is linked not to these "ifs", but to another one further down:

```
%if 1%{!?el8:1}
BUILDTAGS="${BUILDTAGS} no_btrfs"
%endif
```

This "if" expression can only expand to a "true" value, so my understanding is that right now, we always build with flag "no_btrfs", no matter which distro is being used. Assuming that btrfs was never used, I also removed all dependency on btrfs build requires packages.

Romain-Geissler-1A:
Hi,

Are my replies to all concerns ok or I still need to change something ?

Note that RHEL 9 will be officially released this months, so people will start to need the CentOS 9 package ;)

Cheers,
Romain

thaJeztah:
I worked on various permutations of these conditions, and opened https://github.com/docker/containerd-packaging/pull/283, based on your commit; let me know if that looks good to you

PR #271, opened 2022-03-08
Add Ubuntu 22.04.

Romain-Geissler-1A:
Note: Ubuntu 22.04 will be officially released in mid april.


thaJeztah (approved):
LGTM, thanks!

I think we already had this in our internal pipeline (where we set the actual matrix for which we build packages), but it may have been in a testing channel only; I'll check if we can start building packages already (guess it doesn't hurt to be slightly ahead of it being actually released).

Romain-Geissler-1A:
Thanks for merging this. Indeed, if you could start delivering for both Ubuntu 22 and CentOS 9 (we plan on using both in my company, as soon as docker packages are available) that would be cool ! We have been using Ubuntu 22 images already for a few weeks for some of our CI builds and it works fine.

PR #284, opened 2022-06-04
remove deprecated containerd and runc build-tags

thaJeztah:
containerd and runc (no longer) use the seccomp and apparmor build-tags, and runc
has removed the runc_nokmem build-tag (now the default).

seccomp is enabled by default for containerd and runc, but can be disabled on
runc by setting BUILDTAGS to an empty string;
https://github.com/opencontainers/runc/blob/v1.1.2/README.md#build-tags

Given that we always want to include seccomp (with non-static builds), this patch
removes the BUILDTAGS altogether for runc.

For containerd, we still need a buildtag to disable btrfs on CentOS 8 and up;
https://github.com/containerd/containerd/blob/v1.6.5/BUILDING.md#build-containerd

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -36,7 +36,6 @@ binaries: ## Create containerd binaries
 bin/runc:
 	@set -x; GO111MODULE=auto make -C /go/src/github.com/opencontainers/runc  --no-print-directory \
 		BINDIR="$$(pwd)/bin" \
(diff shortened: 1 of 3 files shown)


thaJeztah:
@tianon @rumpl @cpuguy83 PTAL

tianon (approved):

PR #290, opened 2022-08-04
prepare release v1.6.7

thaJeztah:
- Update containerd to v1.6.7
- Update runc to v1.1.3
- Update Golang runtime to 1.17.13 to address CVE-2022-32189

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,16 @@
+containerd.io (1.6.7-1) release; urgency=medium
+
+  * Update containerd to v1.6.7
+  * Update runc to v1.1.3
+  * Update Golang runtime to 1.17.13 to address CVE-2022-32189
+
+ -- Sebastiaan van Stijn <thajeztah@docker.com>  Thu, 04 Aug 2022 22:28:30 +0000
+
 containerd.io (1.6.6-1) release; urgency=high
 
   * Update containerd to v1.6.6 to address CVE-2022-31030
 
- -- Sebastiaan van Stijn <thajeztah@dockercom>  Mon, 06 Jun 2022 20:45:21 +0000
+ -- Sebastiaan van Stijn <thajeztah@docker.com>  Mon, 06 Jun 2022 20:45:21 +0000
 
 containerd.io (1.6.5-1) release; urgency=medium
 
@@ -10,7 +18,7 @@ containerd.io (1.6.5-1) release; urgency=medium
   * Update runc to v1.1.2
   * Update Golang runtime to 1.17.11
 
- -- Sebastiaan van Stijn <thajeztah@dockercom>  Sat, 04 Jun 2022 20:56:32 +0000
+ -- Sebastiaan van Stijn <thajeztah@docker.com>  Sat, 04 Jun 2022 20:56:32 +0000
 
 containerd.io (1.6.4-1) release; urgency=medium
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
@tianon @rumpl ptal

debian/changelog
```diff
@@ -1,16 +1,24 @@
...
+  * Update Golang runtime to 1.17.13 to address CVE-2022-32189
+
+ -- Sebastiaan van Stijn <thajeztah@docker.com>  Thu, 04 Aug 2022 22:28:30 +0000
+
 containerd.io (1.6.6-1) release; urgency=high
 
   * Update containerd to v1.6.6 to address CVE-2022-31030
 
- -- Sebastiaan van Stijn <thajeztah@dockercom>  Mon, 06 Jun 2022 20:45:21 +0000
+ -- Sebastiaan van Stijn <thajeztah@docker.com>  Mon, 06 Jun 2022 20:45:21 +0000
```
thaJeztah, line 13: 🙈 😂 🤷‍♂️

tianon:
"Yep yep, looks like changelog" :joy:
(I'm assuming any actual load-bearing changes for this are already committed prior to this?)

thaJeztah:
Basically "this is it"; well; parameters get passed into the pipeline to build  with the right options, but it needs the changelog 😂 🤷‍♂️ 

(we should perhaps consider generating it during build though)

tianon:
Yeah, I guess the runc bump and the Go bump are the parts I'm surprised not to see here - I know as well as anybody that the version number in the changelog is load-bearing :joy:

thaJeztah:
is that a "LGTM"? 🤗 

I guess I'll use the YOLO power button otherwise

Screenshot 2022-08-05 at 01 08 34

tianon (approved):
Yep!  I won't pretend I know _where_ the updates to `runc` and Go are, but I'll trust you that they're in here.  LGTM :+1:

tianon:
(I'm guessing they're transitively somewhere in the containerd bits, so updating containerd itself is what updates Go and runc)

thaJeztah:
Ah, yes, they are; so we get the default Go version for building from;
https://github.com/containerd/containerd/blob/v1.6.7/contrib/Dockerfile.test#L13

And the runc version from
https://github.com/containerd/containerd/blob/v1.6.7/script/setup/runc-version


Yes, that's a bit "scrappy", but it works nicely. Besides those, we also have variables we can use to explicitly override / customize them if we don't want to follow upstream.

https://github.com/docker/containerd-packaging/blob/655193435b87b50b8b80d81c9a4f550e0f9f3870/scripts/determine-runc-version#L23-L28

https://github.com/docker/containerd-packaging/blob/655193435b87b50b8b80d81c9a4f550e0f9f3870/common/common.mk#L17-L22

thaJeztah:
Getting this one in! (I should add you to the repo to give you a green ✅ 😂

thaJeztah:
oh, fun! _both_ containerd and runc have a all-numbers Git commit (short one that is);

```
Building packages on docker.io/library/debian:bullseye

containerd   : v1.6.7 (commit: 0197261)
INFO: detected runc version (v1.1.3) from script/setup/runc-version
runc         : v1.1.3 (commit: 6724737)
architecture : aarch64
build image  : docker.io/library/debian:bullseye
golang image : golang:1.17.13
```

PR #329, opened 2023-09-26
deb, rpm: use GOTOOLCHAIN=local

thaJeztah:
- relates to https://github.com/docker-library/golang/issues/472

Make sure we don't get unexpected updates of the go toolchain when building.

### DIFF
--- a/dockerfiles/deb.dockerfile
+++ b/dockerfiles/deb.dockerfile
@@ -25,6 +25,7 @@ FROM ${GOLANG_IMAGE} AS golang
 FROM golang AS go-md2man
 ARG GOPROXY=direct
 ARG GO111MODULE=on
+ENV GOTOOLCHAIN=local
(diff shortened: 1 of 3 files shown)


sam-thibault (approved):

rumpl (approved):

thaJeztah:
Looks like Jenkins is failing, but I don't have access (I recall @neersighted was looking into that at some point, but don't know where we ended on)

crazy-max:
> /cc @crazy-max in case similar changes are needed in docker/packaging as well 🤔

Thanks for the ping, will do

thaJeztah:
@crazy-max FWIW; official  images will set this as a default as well (at least in new builds), but if we have parts of the pipeline that are _not_ using `FROM golang`, and (like in this repository) installing `go` in another image, that env-var won't be set. In that case, we need some changes probably  (and set the env-var)

PR #325, opened 2023-09-06
Add fedora 39 (not released yet).

thaJeztah:
- carries / supersedes https://github.com/docker/containerd-packaging/pull/315
- closes https://github.com/docker/containerd-packaging/pull/315

PR #330, opened 2023-10-18
Add Fedora 40 and Ubuntu Noble (not released yet).

Romain-Geissler-1A:
As usual @thaJeztah I submit the pull request, but unless it comes from you the CI won't test anything ;)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ def images = [
     [image: "docker.io/library/debian:bookworm",        arches: ["amd64", "aarch64", "armhf"]], // Debian 12 (stable)
     [image: "docker.io/library/fedora:38",              arches: ["amd64", "aarch64"]],          // Fedora 38 (EOL: May 14, 2024)
     [image: "docker.io/library/fedora:39",              arches: ["amd64", "aarch64"]],          // Fedora 39 (EOL: November 12, 2024)
-    [image: "docker.io/library/fedora:rawhide",         arches: ["amd64", "aarch64"]],          // Rawhide is the name given to the current development version of Fedora
+    [image: "docker.io/library/fedora:40",              arches: ["amd64", "aarch64"]],          // Fedora 40 (EOL: May 13, 2025)
     [image: "docker.io/opensuse/leap:15",               arches: ["amd64"]],
     [image: "docker.io/balenalib/rpi-raspbian:buster",  arches: ["armhf"]],
     [image: "docker.io/balenalib/rpi-raspbian:bullseye",arches: ["armhf"]],
@@ -27,6 +27,7 @@ def images = [
     [image: "docker.io/library/ubuntu:focal",           arches: ["amd64", "aarch64", "armhf"]], // Ubuntu 20.04 LTS (End of support: April, 2025. EOL: April, 2030)
     [image: "docker.io/library/ubuntu:jammy",           arches: ["amd64", "aarch64", "armhf"]], // Ubuntu 22.04 LTS (End of support: April, 2027. EOL: April, 2032)
     [image: "docker.io/library/ubuntu:mantic",          arches: ["amd64", "aarch64", "armhf"]], // Ubuntu 23.10 (EOL: July, 2024)
+    [image: "docker.io/library/ubuntu:noble",           arches: ["amd64", "aarch64", "armhf"]], // Ubuntu 24.04 LTS (End of support: April, 2029. EOL: April, 2034)
 ]
 
 def generatePackageStep(opts, arch) {


Romain-Geissler-1A:
Ping ;)

thaJeztah:
whoops, thought I posted a comment, but looks like I didn't submit.

I don't think this would currently work, because the `fedora:40` image is currently still "rawhide", so it's effectively an alias for `fedora:rawhide`. This was the issue we had with the fedora 39 release as well; https://github.com/docker/containerd-packaging/pull/325#issuecomment-1710737864


```
docker run -it --rm fedora:40
Unable to find image 'fedora:40' locally
40: Pulling from library/fedora
bb33616167e9: Download complete
Digest: sha256:3fd0131e8f99284776d900359f5dbafb2c4611ef086ca56130c1a85fd1653c8d
Status: Downloaded newer image for fedora:40
[root@30d66e6343a1 /]# cat /etc/os-release
NAME="Fedora Linux"
VERSION="40 (Container Image Prerelease)"
ID=fedora
VERSION_ID=40
VERSION_CODENAME=""
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Linux 40 (Container Image Prerelease)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/rawhide/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=rawhide
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=rawhide
SUPPORT_END=2024-05-14
VARIANT="Container Image"
VARIANT_ID=container
```

Until the Fedora project cuts of the release branch for Fedora 40, and publish a beta image, that means packages will be "rawhide", and not a representation of Fedora 40 (or packages may be missing/broken). Looking at the schedule, they target February 6 for that; https://fedorapeople.org/groups/schedule/f-40/f-40-key-tasks.html

Screenshot 2023-11-20 at 23 15 54

Romain-Geissler-1A:
I must be missing some pieces, as I don't understand what the problem actually is.

What I would like to avoid by submitting pull request very soon is the problem we kind of have regularly every 6 months: fedora is out (or almost out), and the Docker packages aren't ready, effectively breaking the people who always pull the latest fedora version.

I get the point that rawhide and the latest fedora version are always an alias. However what I witness is that here https://download.docker.com/linux/fedora/ you currently don't have any package for fedora 40, while technically, it "exist" already, even if not released yet. While now it's not a big problem, the closer we will be to the release date, the more problematic it will be. I think that it's better to build and provide packages for fedora 40 now, and publish it on https://download.docker.com/linux/fedora/, even if it's broken, it's IMO better than having no package at all.

So what is preventing from pushing packages for fedora 40 now (if the build isn't broken) ? If rawhide and 40 are aliases of the same thing, we should drop rawhide, add 40, and publish the packages for 40 at the next release.

Romain-Geissler-1A:
If the concern is about REDHAT_BUGZILLA_PRODUCT_VERSION which currently read rawhide, I don't get why we simply don't use VERSION_ID which has the "right" value.

Romain-Geissler-1A:
I have updated the pull request to include Ubuntu Noble which was published yesterday on Dockerhub. Also I have removed fedora rawhide (in favor of fedora 40, which is the current rawhide).

Romain-Geissler-1A:
Ping ;)

funnelfiasco:
As a Docker employee and long-time Fedora contributor (and former Fedora Program Manager), here are my thoughts:

I view Rawhide as where Fedora package maintainers update and stabilize packages. The beta period is when third-party package maintainers (e.g. us) update and stabilize packages. Publishing packages implies some level of support, which I'm not sure we want to commit to for Fedora Rawhide. Thus, the absolute earliest we should start building for a given Fedora Linux version is the branch date (in the F40 case, that would be 6 February). A better target is the start of the beta freeze (20 February for F40), as that represents a point where the repository churn will be lower and the `fedora:XX` image should exist. That gives us at least 3 weeks, usually 4 or 5, to make sure our packages build before the Fedora Linux beta release.

I'll note that Fedora [provides first-party containerd packages](https://src.fedoraproject.org/rpms/containerd) as well. This will always provide packages for both rawhide and branched releases.

(Also: as a matter of style, it's best to have the Ubuntu and Fedora changes as separate pull requests)

Romain-Geissler-1A:
That makes sense. However my strategy of sending pull request very early is based on past experience: it happens from time to time that while fedora or ubuntu is about to be released, the packages from DockerInc aren't ready. For Fedora 40, if the Fedora release date had not been pushed back couple of times, then the docker package would have been published after Fedora 40, and this scenario happened to me in the past with Fedora/Ubuntu (we use both in my company, and try to adopt stable release as they are shipped).

This plus the fact that containerd and docker do have separate release cycle, that to build the docker package, you first need to publish the containerd package, and that it's impossible to release a package for a new distro without an official containerd/moby release, all these reasons contribute to the fact that on the D day of a distro release, it's not uncommon to be unable to install the DockerInc's packages.

So, if in February, when both Fedora/Ubuntu will be in a beta status you merge this + an equivalent change on docker packaging side + publish new package, fine for me. However my experience is that this usually does not happen like this, that why I try to push this even earlier.

Romain-Geissler-1A:
Hi,

I have rebased this on top of the latest master branch.

Right now Ubuntu noble is in feature freeze, the beta freeze is in a month. Fedora 40 is expected to be in beta freeze in two weeks from what I read (I may be wrong). Rawhide (which you use right now and which this pull request removes) has already moved to fedora 41 so I guess it should be less stable than the current fedora 40.

Is ok to consider merging this now that we are closer to stability ?

thaJeztah:
Merged as;

- https://github.com/docker/containerd-packaging/pull/344
- https://github.com/docker/containerd-packaging/pull/345

PR #348, opened 2024-03-14
debian: add explicit GOARM for raspbian

neersighted:
This is a quick band-aid; we need to overhaul how we handle architectures in these scripts (and use the Debian base images for Raspbian); but for now this should ensure we force ARMv6 for Raspbian targets.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -14,6 +14,14 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+# FIXME: quick hardcoding of GOARM for raspbian; replace with a holistic
+# refactoring of how we handle target architecture
+distribution := $(shell . /etc/os-release; echo "$${ID}")
+ifeq ($(distribution),raspbian)
+	export GOARM=6
+endif
+
+
 %:
 	dh $@ --with systemd
 


thaJeztah (approved):
LGTM

thaJeztah:
Windows failure is unrelated and being worked on.

debian/rules
```diff
@@ -14,6 +14,14 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+# FIXME: quick hardcoding of GOARM for raspbian; replace with a holistic
+# refactoring of how we handle target architecture
+distribution := $(shell . /etc/os-release; echo "$${ID}")
+ifeq ($(distribution),raspbian)
```
tianon, line 20: FWIW, the "Debian native" way to write this would've been including `/usr/share/dpkg/vendor.mk` (or `/usr/share/dpkg/default.mk` which brings in even more useul variables) and checking whether `DEB_VENDOR` is `Raspbian` :+1:

thaJeztah: Oh, that sounds nice. Do you have an example what that looks like? (For me to "code by numbers")? So those files don't go into the `debian/` directory, but elsewhere?

neersighted: They'd go here -- Tianon just means using a makefile `include`. 

Good to know; I cribbed this from other examples of rules that didn't make light of that file 🙈

tianon: Yeah, sorry; more concretely, something like:

```suggestion
include /usr/share/dpkg/vendor.mk
ifeq ($(DEB_VENDOR),Raspbian)
```

tianon: (but as it's temporary, I don't think it really matters all that much)

thaJeztah: Thanks!! ❤️❤️

For sure! Current patch does the job, so I'm ok with it for now. But it's good to learn how to do these things correctly; there's a risk in having "the wrong way" present as it's all too easy to (as one does) taking the "wrong way" as an example and such things starting to spread around.

neersighted: I got the "wrong way" from some of the best maintained packages in Debian, so you're definitely right :joy:

In any case, I think I will have a proposal next week for simplifying this whole mess and putting it semi-permanently out of mind.

PR #352, opened 2024-04-08
prepare release 1.6.31

vvoland:
- update containerd binary to v1.6.31
- Update Golang runtime to 1.21.9

Signed-off-by: Paweł Gronowski <pawel.gronowski@docker.com>

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+containerd.io (1.6.31-1) release; urgency=medium
+
+  * update containerd binary to v1.6.31
+  * Update Golang runtime to 1.21.9
+
+ -- Paweł Gronowski <pawel.gronowski@docker.com>  Wed, 10 Apr 2024 12:46:52 +0000
+
 containerd.io (1.6.28-2) release; urgency=medium
 
   * Fix builds for Raspbian armhf producing armv7 binaries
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,9 @@
+containerd.io (1.6.31-2) release; urgency=medium
+
+  * update containerd binary to v1.6.31.
+
+ -- Paweł Gronowski <pawel.gronowski@docker.com>  Fri, 08 Apr 2024 13:31:20 +0000
```
vvoland, line 5: Should I adjust the date? 🤔

laurazard (dismissed):

debian/changelog
laurazard: I guess? The date matches the commit though 👀

debian/changelog
```diff
@@ -1,3 +1,9 @@
+containerd.io (1.6.31-2) release; urgency=medium
+
+  * update containerd binary to v1.6.31.
```
thaJeztah, line 3: We should build with latest go1.21 version (upstream didn't update yet, but if we build our packages, we can override the version)

thaJeztah (approved):
LGTM, thanks!

PR #359, opened 2024-06-11
Jenkinsfile: switch to Ubuntu 24.04

thaJeztah:


thaJeztah:
Not 100% sure if we have Jenkins agents for 24.04. Let's see 🙈

thaJeztah:
Looks like we don't 😢 - guess we should switch to GitHub actions (there should be arm64 agents now)

PR #361, opened 2024-06-11
[main] minor cleanup now that this branch is for containerd 1.7+

thaJeztah:
Now that the main branch is targeting containerd 1.7, there's some bits we can remove.

### remove GOTOOLCHAIN env-vars

Both containerd and runc have been a module for some time now, and all
the branches we care about are a module, so there's no need to set the
GO111MODULE env-var.

### remove some GOTOOLCHAIN=local env-vars

The official docker image now have GOTOOLCHAIN=local set by default,
so for stages where we build inside that image, we can remove it;

    docker image inspect --format='{{ json .Config.Env }}' golang:1.21.11 | jq .
    [
      "PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "GOLANG_VERSION=1.21.11",
      "GOTOOLCHAIN=local",
      "GOPATH=/go"
    ]

### disable btrfs for all CentOS/RHEL versions

Now that we no longer have to take 1.5/1.6 into account, we must
unconditionally disable brtfs for CentOS and RHEL, as no version
provides the Linux kernel headers for btrfs

### rpm: remove version-constraint for container-selinux

This condition was added 5 Years ago in c08bd94d32f681404b4392ec9f3f94dd6c98ec09
because older versions had some issues.

CentOS 7 currently installs `2:2.119.2-1.911c772.el7_8`, and CentOS 9 stream
has `3:2.231.0-1.el9`; both are higher than the minimum version specified,
so let's remove the condition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,30 @@ make clean
 make docker.io/library/<distro>:<version> [docker.io/library/<distro>:<version> ...]
 
 # for example:
-# make docker.io/library/centos:7
-# make docker.io/library/ubuntu:jammy
+# make quay.io/centos/centos:stream9
+# make docker.io/library/ubuntu:24.04
 ```
 
 After build completes, packages can be found in the `build` directory.
+
+## Specifying the version to build
+
+By default, packages are built from HEAD of the `release/1.7` branch, as
+defines in [common/common.mk]. The version of runc defaults to the version
+as specified by the containerd project through the [script/setup/runc-version]
+file in the containerd repository.
+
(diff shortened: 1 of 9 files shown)


thaJeztah:
This one is ready for review; failures are Fedora Rawhide (unrelated issue)

@dmcgowan @vvoland

dmcgowan (approved):

thaJeztah:
Oh; if you have time; https://github.com/docker/containerd-packaging/pull/362 is also an easy review

PR #362, opened 2024-06-11
[main] Jenkinsfile: remove Fedora 38 as it's EOL

thaJeztah:

PR #363, opened 2024-06-11
prepare release 1.7.18

thaJeztah:
- [x] depends on / stacked on https://github.com/docker/containerd-packaging/pull/361
- [x] depends on / stacked on https://github.com/docker/containerd-packaging/pull/362

This is the first release of containerd.io packages for container 1.7.
Previous packages were using the 1.6 releases, which are marked as
LTS, and would be maintained for longer than the 1.7 release, but
after discussing with containerd maintainers, we learned that the
1.7 release won't reach EOL before 1.6, and thus safe to choose.


thaJeztah:
This change hasn't been completed yet, and isn't critical (but 1.7 would've been a nice "cutover" point to change the package naming scheme); I'll remove this from the change-log;

> This release also updates the package naming scheme to include distro
release version, which helps in situations where the distro is upgraded,
but no newer version of containerd.io is available.

thaJeztah:
And this part also got obsolete now that CentOS 7 / RHEL 7 are reaching EOL, and we stop building packages for those for the upcoming releases;

> This release removes support for the btrfs snapshotter on RHEL 7 and
CentOS 7. Support for btrfs on these platforms was experimental, and
deprecated by the distro.

So removed that part as well

dmcgowan (approved):

PR #373, opened 2024-06-18
update release notes to include runc v1.1.13

thaJeztah:
runc release  v1.1.13, but containerd v1.7.18 does not yet have this update in their "runc-version" file, which is the version we use as default when building;
https://github.com/containerd/containerd/blob/v1.7.18/script/setup/runc-version

So, instead we'll update the RUNC_REF when building packages, so that we can include the current version of runc.

release notes: https://github.com/opencontainers/runc/releases/tag/v1.1.13

update runc to v1.1.13

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,7 @@
 containerd.io (1.7.18-1) release; urgency=medium
 
   * Update containerd binary to v1.7.18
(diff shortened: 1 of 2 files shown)


akerouanton (approved):

thaJeztah:
Thanks!

PR #406, opened 2025-01-10
[main] prepare release 1.7.25

vvoland:
- Update containerd binary to v1.7.25
- Update runc binary to v1.2.4

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+containerd.io (1.7.25-1) release; urgency=medium
+
+  * Update containerd binary to v1.7.25
+  * Update runc binary to v1.2.4
+
+ -- Paweł Gronowski <pawel.gronowski@docker.com>  Fri, 10 Jan 2025 13:41:35 +0000
+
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,10 @@
+containerd.io (1.7.25-1) release; urgency=medium
+
+  * Update containerd binary to v1.7.25
+  * Update runc binary to v1.2.4
```
thaJeztah, line 4: As it's a packaging change, perhaps add a mention of https://github.com/docker/containerd-packaging/pull/403

thaJeztah: Oh! Sorry that's for the RPM packages only, so in the one below 🙈

vvoland: Ah, good point, updated

thaJeztah (approved):
LGTM

thaJeztah:
Sigh; did we remove nodes, or is it just waiting to scale up?

Screenshot 2025-01-10 at 15 36 56

PR #411, opened 2025-03-18
Fix build on debian trixie

vvoland:


dockerfiles/deb.dockerfile
```diff
@@ -51,6 +51,24 @@ RUN apt-get update -q && apt-get install -y --no-install-recommends \
...
+#   collect2: fatal error: cannot find 'ld'
+#
+# Since Trixie, binutils-gold is no longer installed as a part of the binutils package and needs to be installed separately.
+#
+# As a workaround; install binutils-gold if it's not installed
+RUN [ "$(arch)" = 'aarch64' ] && ! command -v ld.gold && apt-get update && apt-get install -y binutils-gold
```
tianon, line 70: `arch` tests the kernel, not the userspace, so this should be something like this instead: 

```suggestion
RUN [ "$(dpkg --print-architecture)" = 'arm64' ] && ! command -v ld.gold && apt-get update && apt-get install -y binutils-gold
```

(Also, it would probably be wise to make this an explicit `if` because otherwise it might not fail the build correctly in all the failure cases)

dockerfiles/rpm.dockerfile
```diff
@@ -74,8 +74,8 @@ FROM redhat-base AS fedora-base
 # - https://src.fedoraproject.org/rpms/golang/blob/a867bd88a656c1d6e91e7b18bab696dc3fcf1e77/f/0006-Default-to-ld.bfd-on-ARM64.patch
 # - https://src.fedoraproject.org/rpms/golang/c/a867bd88a656c1d6e91e7b18bab696dc3fcf1e77?branch=rawhide
 #
-# As a workaround; install binutils-gold
-RUN if [ "$(arch)" = 'aarch64' ] && [ $(source /etc/os-release && echo "${VERSION_ID%.*}") -ge 41 ]; then dnf -y install binutils-gold; fi
+# As a workaround; install binutils-gold if it's not installed
+RUN [ "$(arch)" = 'aarch64' ] && ! command -v ld.gold && dnf -y install binutils-gold
```
tianon, line 78: If you want a userspace alternative for RPM based distros, the best I've seen is `rpm --query --queryformat='%{ARCH}' rpm` (for which `aarch64` *should* be the correct value still).

vvoland:
The remaining failure is in Windows packaging:

```

	git config --global --add safe.directory C:/gopath/src/github.com/containerd/containerd

fatal: detected dubious ownership in repository at 'C:/gopath/src/github.com/containerd/containerd'

'C:/gopath/src/github.com/containerd/containerd' is owned by:

	NT AUTHORITY/SYSTEM (S-1-5-18)

but the current user is:

	User Manager/ContainerAdministrator (S-1-5-93-2-1)

To add an exception for this directory, call:
```

Not sure why it's caused? I don't think we had this error before, and the user should be mapped by the HCS when mapping the host path?
I couldn't reproduce the issue on my local Windows machine.

vvoland:
@thaJeztah are you good with getting this one in? It fixes the Debian Trixie failure and the remaining one is Windows.

thaJeztah (approved):
LGTM

PR #412, opened 2025-03-18
Drop Windows packaging

vvoland:
- needs: https://github.com/docker/containerd-packaging/pull/411

The Windows stage is currently broken (https://github.com/docker/containerd-packaging/pull/411#issuecomment-2733007722)

We don't publish the Windows binaries anyway, so let's just drop the packaging for Windows.

If there's a need for us to ship Windows containerd binaries (e.g. to use in Docker Desktop), we should be able to just use upstream binaries.

Signed-off-by: Paweł Gronowski <pawel.gronowski@docker.com>


thaJeztah:
We should probably look at the reverse, and see if we can make the Windows packaging work (also because we want to move the engine to use containerd as runtime for windows).

What's currently broken? Was that something that broke recently?

thaJeztah:
I think it's good practice to have packages for things we ship in a single, trusted location, and not require users to hunt down requirements from various repositories on GitHub. It also makes sure that we keep the versions aligned with what we ship for Linux.

thaJeztah:
So the windows failure looks possibly due to an updated version of Git being more restrictive? We're bind-mounting the code from the host, but the container runs with the default (`ContainerAdministrator`) user, and it complains ownership of the files doesn't match.

I wonder if that's something that users will start running into if that's indeed a new change in Git;

```
docker run \
	--rm \
	-v "d:/jenkins/workspace/containerd-packaging_PR-411/src/:C:/gopath/src" \
	-v "d:/jenkins/workspace/containerd-packaging_PR-411/build/windows:C:/gopath/src/github.com/containerd/containerd/bin" \
	-w "C:/gopath/src/github.com/containerd/containerd" \
	dockereng/containerd-windows-builder \
	make bin/containerd
fatal: detected dubious ownership in repository at 'C:/gopath/src/github.com/containerd/containerd'
'C:/gopath/src/github.com/containerd/containerd' is owned by:
	NT AUTHORITY/SYSTEM (S-1-5-18)
but the current user is:
	User Manager/ContainerAdministrator (S-1-5-93-2-1)
To add an exception for this directory, call:

	git config --global --add safe.directory C:/gopath/src/github.com/containerd/containerd
fatal: detected dubious ownership in repository at 'C:/gopath/src/github.com/containerd/containerd'
'C:/gopath/src/github.com/containerd/containerd' is owned by:
	NT AUTHORITY/SYSTEM (S-1-5-18)
but the current user is:
	User Manager/ContainerAdministrator (S-1-5-93-2-1)
To add an exception for this directory, call:
```

thaJeztah:
Oh! I just recalled we had something with this before on Linux; I'm wondering if the machine we're running on got updated and/or the Windows image used. There's a workaround for that to set the path as "save path";

- https://github.com/moby/moby/pull/44930

thaJeztah:
Oh, LOL, but that's the option we're trying to set there? 🤔 weird. Perhaps because the git repository is already there and something changed in Git again?

I see others used a wildcard for this, perhaps that works 🤔 https://github.com/grpc/grpc/blob/8cc807bc71b4e642488b0e4c09e3f76596b6fe28/tools/dockerfile/interoptest/grpc_interop_nodepurejs/Dockerfile#L57-L62

vvoland:
Yeah I tried it with the same git on my local Windows and couldn't reproduce this. Anyway, I'd like to keep this PR as a proposal to just remove the Windows packaging, unrelated to the actual Windows failure 😄

vvoland:
> I think it's good practice to have packages for things we ship in a single, trusted location, and not require users to hunt down requirements from various repositories on GitHub. It also makes sure that we keep the versions aligned with what we ship for Linux.

We'd still package it, we'd just use the upstream containerd .exe instead of building them ourselves

thaJeztah:

PR #414, opened 2025-03-18
Fix Windows packaging

vvoland:
Adds the containerd path to safe directories to address the "dubious
ownership" error when building containerd in Windows container:

```
fatal: detected dubious ownership in repository at 'C:/gopath/src/github.com/containerd/containerd'

'C:/gopath/src/github.com/containerd/containerd' is owned by:

	NT AUTHORITY/SYSTEM (S-1-5-18)

but the current user is:

	User Manager/ContainerAdministrator (S-1-5-93-2-1)
```

PR #415, opened 2025-03-20
Jenkinsfile: Temporarily disable Windows build

vvoland:
Windows is currently broken: https://github.com/docker/containerd-packaging/pull/414

Since we're not shipping Windows binaries anyway, let's disable it for now to allow us publishing the Linux packages.


Jenkinsfile
```diff
@@ -67,22 +67,7 @@ def generatePackageSteps(opts) {
...
-def packageBuildSteps = [
-    "windows": { ->
-        node("windows-2022") {
-            stage("windows") {
-                try {
-                    checkout scm
-                    sh("make -f Makefile.win archive")
-                } finally {
-                    deleteDir()
-                }
```
thaJeztah, line 79: Should we catch the error, so that we still have information about the failures while we're investigating? Not exactly sure what the right format is, but this SO post has some examples
https://stackoverflow.com/questions/44022775/jenkins-ignore-failure-in-pipeline-build-step

vvoland: Hmmm good one, will try it out

thaJeztah (approved):
LGTM, thanks!

PR #416, opened 2025-03-21
Fix shell script vulnerabilities

sundanc:
**- What I did**

Fixed two shell script vulnerabilities that could lead to unexpected behavior.

**- How I did it**

- Added the missing `pipefail` argument to the `-o` flag in `scripts/validate/fileheader`
- Changed `checkout $*` to `checkout "$@"` in `scripts/checkout.sh` to properly handle arguments with spaces

**- How to verify it**

- Test the validate script to ensure it behaves correctly on pipeline failures
- Test the checkout script with arguments containing spaces to verify they are handled correctly

Fixed shell script vulnerabilities in validate/fileheader and checkout.sh scripts.

### DIFF
--- a/scripts/validate/fileheader
+++ b/scripts/validate/fileheader
@@ -14,7 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-set -eu -o
+set -eu -o pipefail
 
 if ! command -v ltag; then
 	echo >&2 "ERROR: ltag not found. Install with:"
--- a/scripts/checkout.sh
+++ b/scripts/checkout.sh
@@ -56,5 +56,5 @@ checkout() (
 # Only execute checkout function above if this file is executed, not sourced from another script
 prog=checkout.sh # needs to be in sync with this file's name
 if [ "$(basename -- $0)" = "$prog" ]; then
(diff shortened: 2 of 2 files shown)


scripts/validate/fileheader
```diff
@@ -14,7 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-set -eu -o
+set -eu -o pipefail
```
thaJeztah, line 17: We should probably check if it's already supported on all distros; this script uses a POSIX shell (not Bash), and  `pipefail` was not part of POSIX until recently; some distros (specifically; Debian) may not support it yet; also see

- https://github.com/koalaman/shellcheck/issues/2555

tianon: Either way, `set -eu -o` without an argument to that `-o` is wrong, so this should be updated even in master (one way or the other) :sweat_smile:

thaJeztah: Ah! Yes! I blame ..... me for that (hehe and look at that diff 🙈🥹) https://github.com/docker/containerd-packaging/commit/93ae009441671c3e3f0488ae2890644211c99d8a

Issue #303, opened 2023-01-02
containerd.io package naming is missing the distribution


devopstt:

tirolerstefan:
Hi @thaJeztah - thanks for linking my issue  "Different checksum..." to this project.
Do you roughly know when this could be implemented? Thanks!

marxarelli:
We just ran into this as well with our reprepro configuration. It would be great to have a fix!

thaJeztah:
closing, as we moved our release pipeline to use https://github.com/docker/packaging;

- https://github.com/docker/containerd-packaging/pull/429

And packages built from that pipeline now have a proper suffix;


<img width="501" height="182" alt="Image" src="https://github.com/user-attachments/assets/91e3d3de-582e-4f76-a205-305bcfaa6c30" />

PR #423, opened 2025-05-28
Use SOURCE_DATE_EPOCH to improve reproducibility

marxarelli:
Set `SOURCE_DATE_EPOCH` to the most recent modification timestamp of all repos under `src/`. This allows deb and rpm toolchains to clamp mtimes and better reproduce packages that would otherwise be different due solely to the current time.

See https://reproducible-builds.org/docs/source-date-epoch/ for a recommended approaches to setting this variable and a list of supported tooling.

Note that this was developed while testing a fix to #303 but it does not address that issue directly as compiled binaries still differ between Debian releases.

**- What I did**

Introduced a `SOURCE_DATE_EPOCH` build argument to make package builds more reproducible. Support varies by package build tool.

**- How I did it**

Defined a new `SOURCE_DATE_EPOCH` build arg and set it to the most recent modification timestamp of all respos under `src/`. 

**- How to verify it**

 1. Start with a fresh build cache (e.g. `docker buildx prune`) and build directory (`rm -rf build/*`).
 2. Build the trixie deb with `make REF=v1.7.27 PROGRESS=plain ARCH=amd64 docker.io/library/debian:trixie`
 3. Copy the results to a temporary location (e.g. `cp -a build build1`).
 4. Repeat steps 1 and 2 to clear the cache and build directory again and rebuild the packages.
 5. Compare the two build directories (`diff -r build build1`). They should be identical.

Improved package reproducibility by clamping mtimes with `SOURCE_DATE_EPOCH`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,12 @@ BUILD_BASE=$(shell ./scripts/determine-base $(BUILD_IMAGE))
 PROGRESS=auto
 TARGET=packages
 
+# SOURCE_DATE_EPOCH is set to make builds reproducible that would otherwise be
+# different due solely to file mtimes and derived changelog entries. It is set
+# to the most recent modification timestamp of all repos under src/. The rpm
+# and deb packaging toolchains support this variable.
+SOURCE_DATE_EPOCH=$(shell ./scripts/source-date-epoch src/)
+
 all: build
 
 .PHONY: clean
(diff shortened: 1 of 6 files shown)


marxarelli:
@thaJeztah not sure if you all want this but I thought I would submit it just in case.

I plan to follow up with a PR that addresses #303 directly by including the Debian release name/number in the package version string.

vvoland:
Hi, this repository is no longer used for containerd packaging by Docker Inc.

Would you be interested in porting this change to https://github.com/docker/packaging/tree/main/pkg/containerd?

vvoland:
Thanks for the contribution! I'm going to close this one as the repo is defunct.

